PR #2, opened 2021-04-01
add battle's views and models to review

DeborahAlmeida:
## Objective and Motivation
- As a Trainer, I create a new battle
- As the opponent Trainer I select my PKNs to battle
- As a Trainer I see form errors in case the PKNs I select sum more than 600 points


rvlb (commented):
Tested the flow, worked fine, congrats.

The PR has a conflict that was probably caused by deleting the `.pyc` on `master`.

PR #3, opened 2021-04-07
send battle's result by e-mail

DeborahAlmeida:
## Objective and Motivation
- Send battle's result to players by e-mail


rvlb (approved):
Overall looking good.

2 things to look out for:
1. On a future PR it could be worth refactoring the validation logic on `round_new` and `round_new2` and put it on a new function (like `validate_pokemon_team(pkmn1, pkmn2, pkmn3)`, to avoid repeating code
2. There are some files (db.sqlite3, .pyc files) that are appearing here even though they are on .gitignore. Maybe it's because the gitignore was added midway. Although commiting these types of files doesn't cause problems, it's recommended that they're a kept outside git version control.

DeborahAlmeida:
Thank you! I'm going to refactor my view's code on the next task. I appreciate your suggestions! :)

PR #6, opened 2021-05-04
Feature/migrations models

DeborahAlmeida:
## Objective and Motivation
- These are the right changes, keeping the old migrations and creating new ones to the actual model


backend/battle/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 3.2 on 2021-04-28 18:39
+# Generated by Django 2.2.19 on 2021-03-22 05:52
```
rsarai: Apparently you have downgraded Django, any specific reason for that?

rsarai: Is this an older migration that you are bringing it back?

backend/pokemon/models.py
```diff
@@ -1,12 +1,10 @@
-from django.conf import settings
+# from django.conf import settings
 from django.db import models
 
+
 class Pokemon(models.Model):
-    id = models.AutoField(primary_key=True)
+    pokemon_id = models.CharField(max_length=200, null=True)
```
rsarai: Why do you need an id saved as char?

rsarai (changes requested):

backend/battle/migrations/0001_initial.py
DeborahAlmeida: I didn't realize that I have downgraded Django, I'm gonna fix this

DeborahAlmeida: About this migration, yeah, I brought it again, because when I was resolving conflicts, I deleted it.

backend/pokemon/models.py
DeborahAlmeida: Actually, I don't need it, I will resolve this saving as Integer.

rsarai (approved):

PR #8, opened 2021-05-14
create view to add battle

DeborahAlmeida:
## Objective and Motivation
- Create a battle using class-based views
- Create a team
- Create pokemon's team


backend/battle/forms.py
```diff
@@ -1,14 +1,47 @@
...
+from django import forms
+
+from .models import Battle, PokemonTeam, Team
+from pokemon.models import Pokemon
+
+class TeamForm(forms.ModelForm):
+    class Meta:
+        model = PokemonTeam
+        fields = [
+        'pokemon_1',
```
rsarai: Identation issue here :see_no_evil:

backend/battle/forms.py
```diff
@@ -1,14 +1,47 @@
...
+        'pokemon_2',
+        'pokemon_3'
+        ]
+
+    pokemon_1 = forms.IntegerField(
```
rsarai: Please use `ModelMultipleChoiceField` or  `ModelChoiceField`: https://docs.djangoproject.com/en/3.2/topics/forms/modelforms/#field-types

backend/battle/forms.py
```diff
@@ -1,14 +1,47 @@
...
+    #     PokemonTeam.objects.create(team=Team.objects.get(pk=47), pokemon=Pokemon.objects.get(pokemon_id=18), order=1)
+    #     PokemonTeam.objects.create(team=Team.objects.get(pk=47), pokemon=Pokemon.objects.get(pokemon_id=16), order=2)
+    #     PokemonTeam.objects.create(team=Team.objects.get(pk=47), pokemon=Pokemon.objects.get(pokemon_id=15), order=3)
+    #     instance = super().save(commit=False)
+    #     instance.some_flag = True
+    #     if commit:
+    #         instance.save()
+    #         self.save_m2m()
+
+    #     return instance
```
rsarai: why the comments?

backend/battle/urls.py
```diff
@@ -1,18 +1,15 @@
...
 urlpatterns = [
     path("", Home.as_view(), name="home"),
-    # path('round/new/', RoundNewCreator.as_view(), name='round_new'),
-    # path('invite/', Invite.as_view(), name='invite'),
-    # path('opponent/', Opponent.as_view(), name='opponent'),
-    # path('player2/round', RoundNewOponnent.as_view() , name='round_new2'),
-
+    path("battle/", BattleView.as_view(), name="battle"),
+    path('invite/', Invite.as_view(), name='invite'),
+    path("team/<int:pk>/create", TeamView.as_view(), name="team_create"),
```
rsarai: The pk here is related to the team or the battle?

backend/battle/views.py
```diff
@@ -9,12 +13,51 @@
...
 
-    def get(self, request):
-        return render(request, self.template_name)
 
+class Invite(TemplateView):
+    template_name = 'battle/invite.html'
+
+
+class BattleView(CreateView):
+    model = Battle
```
rsarai, line 21: which form is being used here?

backend/battle/views.py
```diff
@@ -9,12 +13,51 @@
...
+class TeamView(CreateView):
+    template_name = "battle/pokemon_form.html"
+    form_class = TeamForm
+    success_url = reverse_lazy("home")
+
+    def get(self, request, pk):
+        self.object = None
+        battle_object = Battle.objects.get(pk=pk)        
+        Team.objects.create(battle=battle_object, trainer=battle_object.creator)
+        Team.objects.create(battle=battle_object, trainer=battle_object.opponent)
```
rsarai: So every time a user access the page new teams will be created? Is this the correct behavior?

backend/battle/views.py
```diff
@@ -9,12 +13,51 @@
...
+        return super().get(request, pk)
+
+
+    # def form_invalid(self, form):
```
rsarai: Please let's try to avoid adding new code with comments

backend/battle/views.py
```diff
@@ -9,12 +13,51 @@
...
+    model = Battle
+    fields = ['opponent']
+
+    def form_valid(self, form):
+        form.instance.creator = self.request.user
+        form.save()
+        return HttpResponseRedirect(reverse_lazy("team_create", args=(form.instance.id,)))
+
+class TeamView(CreateView):
+    template_name = "battle/pokemon_form.html"
```
rsarai, line 37: To which model is this view related?

rsarai (changes requested):

backend/battle/views.py
DeborahAlmeida: To Team model, I will fix this

backend/battle/views.py
DeborahAlmeida: Sorry about it, I'm gonna delete what is not necessary

backend/battle/views.py
DeborahAlmeida: That's wrong, I will think about a new solution. Thanks :)

backend/battle/views.py
DeborahAlmeida: I didn't make a form for this view, because I thought the CreateView creates a register without a form. I will create a form for this.

backend/battle/urls.py
DeborahAlmeida: It's related to the battle

backend/battle/forms.py
DeborahAlmeida: I was afraid to lose what I wrote, but this is totally wrong. I'm gonna delete it.

backend/battle/forms.py
DeborahAlmeida: Sure, thanks!

backend/battle/forms.py
DeborahAlmeida: Ops, thanks! I'm gonna fix it :D

backend/battle/admin.py
```diff
@@ -1,6 +1,16 @@
...
+    extra = 1
+
+
+class TeamAdmin(admin.ModelAdmin):
+    inlines = (PokemonsInline,)
+
+
 admin.site.register(Battle)
-admin.site.register(Team)
+admin.site.register(Team, TeamAdmin)
```
rsarai, line 15: Nice!

backend/battle/forms.py
```diff
@@ -1,24 +1,57 @@
...
+        queryset=Pokemon.objects.all(),
+        required=True,
+    )
+    pokemon_3 = forms.ModelChoiceField(
+        label="Pokemon 3",
+        queryset=Pokemon.objects.all(),
+        required=True,
+    )
+
+    def __init__(self, *args, **kwargs):
```
rsarai, line 52: is this required?

backend/battle/forms.py
```diff
@@ -1,24 +1,57 @@
...
+        super().__init__(*args, **kwargs)
+
+    def clean(self):
```
rsarai, line 57: A good validation here should be to check if the trainer is or the opponent or the creator of the battle

backend/battle/urls.py
```diff
@@ -1,18 +1,20 @@
...
 
 urlpatterns = [
     path("", Home.as_view(), name="home"),
-    # path('round/new/', RoundNewCreator.as_view(), name='round_new'),
-    # path('invite/', Invite.as_view(), name='invite'),
-    # path('opponent/', Opponent.as_view(), name='opponent'),
-    # path('player2/round', RoundNewOponnent.as_view() , name='round_new2'),
+    path("battle/", BattleView.as_view(), name="battle"),
+    path('invite/', Invite.as_view(), name='invite'),
+    path('battle/list', BattleList.as_view(), name='battle_list'),
```
rsarai: URLs must finish with `/`

backend/battle/views.py
```diff
@@ -1,98 +1,57 @@
...
-# from .battles.base_stats import get_total_point_pokemon
-
-
-class Home(View):
+from django.http import HttpResponseRedirect
+from django.db.models import Q
+from django.views.generic import TemplateView, CreateView, ListView, DetailView
+from .models import Battle, Team
+from django.urls import reverse_lazy
+from .forms import TeamForm, BattleForm
```
rsarai: I try to avoid relative imports, it can cause problems when copying and pasting code or when changing files location

backend/battle/urls.py
```diff
@@ -1,18 +1,20 @@
...
+    path("team/<int:pk>/create/<int:user>", TeamView.as_view(), name="team_create"),
```
rsarai: You don't need to pass the user here. You can use the logged user inside the view

backend/battle/views.py
```diff
@@ -1,98 +1,57 @@
...
+
+class BattleView(CreateView):
+    model = Battle
+    template_name = 'battle/battle_form.html'
+    form_class = BattleForm
+
+    def form_valid(self, form):
+        form.instance.creator = self.request.user
+        form.save()
+        return HttpResponseRedirect(reverse_lazy("team_create", args=(form.instance.id, 1, )))
```
rsarai: I suggested to change the url params by removing the user's id

backend/battle/views.py
```diff
@@ -1,98 +1,57 @@
...
+
+
+class TeamView(CreateView):
+    model = Team
+    template_name = "battle/pokemon_form.html"
+    form_class = TeamForm
+    success_url = reverse_lazy("invite")
+
+    def get_initial(self):
+        super(TeamView, self).get_initial()
```
rsarai: since `get_initial` returns an dict you should do: `self.initial = super(TeamView, self).get_initial()`

backend/battle/views.py
```diff
@@ -1,98 +1,57 @@
...
+        self.initial = {"battle": self.kwargs['pk'], "user": self.kwargs['user']}
```
rsarai: The initial fields should be the same ones as the form. In this case, `trainer` and `battle`.

rsarai: Also, instead of using the `self.kwargs['pk']` you can pass the battle object.

backend/battle/forms.py
```diff
@@ -1,24 +1,57 @@
...
+        cleaned_data = super().clean()
+        return cleaned_data
+
+    def save(self, commit=True):
+        battle_object = Battle.objects.get(pk=self.initial['battle'])
+        if self.initial['user'] == 1:
```
rsarai: At this point, the user passed to the form should be the correct user. The validations on `clean()` should ensure that

backend/battle/forms.py
```diff
@@ -1,24 +1,57 @@
...
+            team = Team.objects.create(battle=battle_object, trainer=battle_object.creator)
+        else:
+            team = Team.objects.create(battle=battle_object, trainer=battle_object.opponent)
+        PokemonTeam.objects.create(team=team, pokemon=Pokemon.objects.get(pokemon_id=18), order=1)
+        PokemonTeam.objects.create(team=team, pokemon=Pokemon.objects.get(pokemon_id=16), order=2)
+        PokemonTeam.objects.create(team=team, pokemon=Pokemon.objects.get(pokemon_id=15), order=3)
```
rsarai: You should use the data selected from the `pokemon_1`, `pokemon_2`, and `pokemon_3` fields.

rsarai (changes requested):

backend/battle/forms.py
DeborahAlmeida: It's not, I think I overwrote without use

backend/battle/forms.py
DeborahAlmeida: I do this validation on def save() but I think it's not the right place to make this. I'm gonna make on def clean()

backend/battle/urls.py
DeborahAlmeida: thanks!

backend/battle/views.py
DeborahAlmeida: thank you!

backend/battle/views.py
DeborahAlmeida: Okay :)

backend/battle/views.py
DeborahAlmeida: hmm I see

backend/battle/views.py
DeborahAlmeida: thanks :D

backend/battle/battles/battle.py
```diff
@@ -1,29 +1,47 @@
...
+def battleRunning(team):
+    battle_info = team.battle
+    pokemon_team_opponent = PokemonTeam.objects.filter(team=team)
+    team_creator = Team.objects.get(battle=battle_info, trainer=battle_info.creator)
+    pokemon_team_creator = PokemonTeam.objects.filter(team=team_creator)
+    winner = get_points(pokemon_team_opponent, pokemon_team_creator)
+    return winner
+
+
+def get_points(pokemons_creator, pokemons_opponent):
```
rsarai: I think the name here is misleading, you are not getting the points, you are fetching the winner.

backend/battle/battles/battle.py
```diff
@@ -1,29 +1,47 @@
...
-# def message_error():
-#     message = "ERROR: PKNs you selected sum more than 600 points, please choose again"
-#     return message
+from battle.models import Battle
+from battle.battles.rounds import roundRunning
+from battle.models import PokemonTeam, Team
+from battle.battles.email import result_battle
+
+
+def battleRunning(team):
```
rsarai: method names should be `snake_case`

backend/battle/battles/battle.py
```diff
@@ -1,29 +1,47 @@
...
+    pokemon_team_creator = PokemonTeam.objects.filter(team=team_creator)
+    winner = get_points(pokemon_team_opponent, pokemon_team_creator)
+    return winner
+
+
+def get_points(pokemons_creator, pokemons_opponent):
+    battle_rounds = [
+        [pokemons_creator[0].pokemon, pokemons_opponent[0].pokemon],
+        [pokemons_creator[1].pokemon, pokemons_opponent[1].pokemon],
+    ]
```
rsarai: I'm confused, shouldn't this be three pokemon? Also, you can pass the teams and use `team_creator.pokemons` and `team_opponent.pokemons`

backend/battle/battles/battle.py
```diff
@@ -1,29 +1,47 @@
...
+
+def validate_sum_pokemons(pokemons):
+    total_points = 0
+    for pokemon in pokemons:
+        pokemon_point = pokemon.attack + pokemon.defense + pokemon.hp
+        total_points += pokemon_point
+    return total_points <= 600
+
+
+def setWinner(winner, battle):
```
rsarai: method names should be `snake_case`

backend/battle/forms.py
```diff
@@ -1,24 +1,76 @@
...
 
+class BattleForm(forms.ModelForm):
+    class Meta:
+        model = Battle
+        fields = ['opponent', ]
 
-# class BattleForm(forms.ModelForm):
+    def clean(self):
+        cleaned_data = super().clean()
+        if cleaned_data['opponent'] == self.initial['creator']:
```
rsarai: Why not do the following?
```python
if cleaned_data['opponent'] == self.cleaned_data['creator']:
```

backend/battle/forms.py
```diff
@@ -1,24 +1,76 @@
...
+                cleaned_data['pokemon_1'],
+                cleaned_data['pokemon_2'],
+                cleaned_data['pokemon_3']
+            ]
+        )
+        obj_battle = cleaned_data['battle']
 
-# class RoundForm2(forms.ModelForm):
+        if cleaned_data['trainer'] != obj_battle.creator:
+            if cleaned_data['trainer'] != obj_battle.opponent:
```
rsarai: This could be replaced by:

```python
if cleaned_data['trainer'] != obj_battle.creator and cleaned_data['trainer'] != obj_battle.opponent:

```

backend/battle/templates/battle/battle_list.html
```diff
@@ -0,0 +1,109 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>Pokebattle</title>
+  <style>
+    body{
+        padding: 0px;
+        background: #1488CC;  /* fallback for old browsers */
+        font-family: Arial, Helvetica, sans-serif;
+
```
rsarai: extra spaces here

backend/battle/views.py
```diff
@@ -1,98 +1,73 @@
...
+    template_name = 'battle/invite.html'
+
+
+class BattleView(CreateView):
+    model = Battle
+    template_name = 'battle/battle_form.html'
+    form_class = BattleForm
+
+    def get_initial(self):
+        obj_creator = get_object_or_404(User, email=self.request.user)
```
rsarai: This is not required, the logged user is already saved on the request, you can just do: `self.initial = {"creator": self.request.user}`

backend/battle/views.py
```diff
@@ -1,98 +1,73 @@
...
+        self.initial = {"creator": obj_creator}
+        return self.initial
+
+    def form_valid(self, form):
+        form.instance.creator = self.request.user
```
rsarai: is this required? Since you are already passing on the `get_initial`

backend/battle/views.py
```diff
@@ -1,98 +1,73 @@
...
+
+class TeamView(CreateView):
+    model = Team
+    template_name = "battle/pokemon_form.html"
+    form_class = TeamForm
+    success_url = reverse_lazy("invite")
+
+    def get_initial(self):
+        obj_battle = get_object_or_404(Battle, pk=self.kwargs['pk'])
+        obj_trainer = get_object_or_404(User, email=self.request.user)
```
rsarai: Same as above, you don't need to fetch the user since it's already loaded on the request

backend/battle/views.py
```diff
@@ -1,98 +1,73 @@
...
+
+    def form_valid(self, form):
+        team = form.save()
+        if team.battle.creator == team.trainer:
+            return HttpResponseRedirect(reverse_lazy("invite"))
+        if team.battle.opponent == team.trainer:
+            result = battleRunning(team)
+            setWinner(result, team.battle)
+            return HttpResponseRedirect(reverse_lazy("home"))
+        return True
```
rsarai: This redirect logic is a bit misleading. What happens when the opponent creates a team before the creator? I think you should only run the battle if both teams are created.

rsarai (changes requested):

backend/battle/forms.py
DeborahAlmeida: It's because I don't have the creator value on cleaned_data by default. I put the creator value on form_valid(). So to solve this I pass the self.request.user by get_initial()

backend/battle/views.py
DeborahAlmeida: Nop :)

backend/battle/views.py
DeborahAlmeida: But when I send the self.request.user, it gives me an error because it's not the object, how can I solve this?

backend/battle/forms.py
DeborahAlmeida: I fixed it :) thanks

backend/battle/forms.py
DeborahAlmeida: Oh great! Thank you :)

backend/battle/battles/battle.py
```diff
@@ -1,29 +1,42 @@
...
-# def message_error():
-#     message = "ERROR: PKNs you selected sum more than 600 points, please choose again"
-#     return message
+from battle.battles.rounds import run_round
+from battle.battles.email import result_battle
+
+
+def run_battle(battle):
+    teams = battle.teams.all()
+    assert teams.count() == 2, "Unexpected error. Battle has more than two teams."
```
fjsj: Just a small optimization you may want to learn here.
If you do `battle.teams.all()` then `(teams[0], teams[1])` you do 1 DB query. But `teams.count()` does another DB query.

Since you're going to fetch anyway the team items (0 and 1), you can always ensure 1 DB query only by doing this:

```
teams = list(battle.teams.all())
assert len(teams) == 2, "Unexpected error. Battle has more than two teams."
team_winner = get_winner_for(teams[0], teams[1])
```

backend/battle/battles/email.py
```diff
@@ -1,20 +1,13 @@
...
     send_templated_mail(
         template_name='results_battle',
         from_email='deborahmendonca6@gmail.com',
-        recipient_list=['deborah.mendonca@vinta.com.br'],
+        recipient_list=[battle.creator, battle.opponent],
         context={
-            'winner': winner,
+            'winner': battle.winner,
         },
         # Optional:
```
fjsj: Please remove those comments

backend/battle/models.py
```diff
@@ -24,7 +24,7 @@ class Team(models.Model):
 class PokemonTeam(models.Model):
     team = models.ForeignKey(Team, on_delete=models.CASCADE, related_name="teams")
     pokemon = models.ForeignKey(Pokemon, on_delete=models.CASCADE, related_name="+")
-    order = models.PositiveIntegerField()
+    order = models.PositiveIntegerField(null=True)
```
fjsj: Sorry if you already discussed this with Rebeca, but does this really need to be `null=True`?

backend/battle/views.py
```diff
@@ -1,98 +1,71 @@
...
+    template_name = 'battle/invite.html'
+
+
+class BattleView(CreateView):
+    model = Battle
+    template_name = 'battle/battle_form.html'
+    form_class = BattleForm
+
+    def get_initial(self):
+        obj_creator = get_object_or_404(User, email=self.request.user)
```
fjsj: I think this is still wrong, no? `obj_creator = self.request.user` doesn't work?

backend/pokemon/models.py
```diff
@@ -5,6 +5,6 @@
 class Pokemon(models.Model):
     pokemon_id = models.IntegerField(null=True)
     name = models.CharField(max_length=200, null=True)
-    attack = models.CharField(max_length=200, null=True)
-    defense = models.CharField(max_length=200, null=True)
-    hp = models.CharField(max_length=200, null=True)
+    attack = models.IntegerField(null=True)
+    defense = models.IntegerField(null=True)
+    hp = models.IntegerField(null=True)
```
fjsj: Sorry if you already discussed this with Rebeca, but does all fields here need to be `null=True`?

fjsj (changes requested):
Please check comments

backend/battle/battles/battle.py
DeborahAlmeida: Great! I didn't know about it. I'm going to improve this :) Thank you

backend/battle/battles/email.py
DeborahAlmeida: Okay, thank you for remember me :)

backend/battle/models.py
DeborahAlmeida: Previously it was a requirement because I used to create a register on PokemonTeam without order and then I updated it. But now it's not like this, so I'm going to change it. Thank you one more time :)

backend/battle/views.py
DeborahAlmeida: Actually, it works fine, I didn't realize any error with this. Maybe it has something I didn't recognize. Can you explain why this could cause an error?

backend/pokemon/models.py
DeborahAlmeida: Actually, they don't. I'll change this. Thanks!

backend/battle/views.py
fjsj: maybe you did get an error when you accessed this page without logging in beforehand.
but either way, it would fail without this as well

DeborahAlmeida: thank you for the explanation!

rsarai (approved):

PR #9, opened 2021-06-08
list battles by status

DeborahAlmeida:
## Objective and Motivation
- List all battles related to user
- List by status: Settled and On going

### DIFF
--- a/backend/battle/views.py
+++ b/backend/battle/views.py
@@ -70,11 +70,6 @@ class BattleDetail(DetailView):
     model = Battle
     template_name = "battle/battle_detail.html"
 
-    def get(self, request, *args, **kwargs):
-        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
-        context = self.get_context_data(object=self.object)
-        return self.render_to_response(context)
-
     def get_context_data(self, **kwargs):
         context = {}
         if self.object:
--- a/backend/battle/views.py
+++ b/backend/battle/views.py
@@ -72,6 +72,17 @@ class BattleDetail(DetailView):
 
     def get(self, request, *args, **kwargs):
         self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
-        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
-        context = self.get_context_data(object=self.object, team=team)
+        context = self.get_context_data(object=self.object)
         return self.render_to_response(context)
+
+    def get_context_data(self, **kwargs):
+        context = {}
+        if self.object:
+            context['object'] = self.object
+            context_object_name = self.get_context_object_name(self.object)
+            if context_object_name:
+                context[context_object_name] = self.object
+        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
+        kwargs['team'] = team
+        context.update(kwargs)
+        return super().get_context_data(**context)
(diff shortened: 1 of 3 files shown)


backend/battle/views.py
```diff
@@ -69,3 +69,9 @@ def get_queryset(self):
 class BattleDetail(DetailView):
     model = Battle
     template_name = "battle/battle_detail.html"
+
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
+        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
+        context = self.get_context_data(object=self.object, team=team)
+        return self.render_to_response(context)
```
rsarai: Please update the context on the `get_context_data`

DeborahAlmeida: Okay, thanks!

backend/battle/views.py
```diff
@@ -69,3 +69,20 @@ def get_queryset(self):
 class BattleDetail(DetailView):
     model = Battle
     template_name = "battle/battle_detail.html"
+
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()  # pylint: disable=attribute-defined-outside-init
+        context = self.get_context_data(object=self.object)
+        return self.render_to_response(context)
```
fjsj: Why do you need to override this `def get`?

backend/battle/views.py
```diff
@@ -69,3 +69,20 @@ def get_queryset(self):
...
+
+    def get_context_data(self, **kwargs):
+        context = {}
+        if self.object:
+            context['object'] = self.object
+            context_object_name = self.get_context_object_name(self.object)
+            if context_object_name:
+                context[context_object_name] = self.object
```
fjsj: I seems those lines belong to the `DetailView`. You can get this behavior by doing `context = super().get_context_data(**kwargs)`, thereby avoiding copying this code.

backend/battle/views.py
```diff
@@ -69,3 +69,20 @@ def get_queryset(self):
...
+        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
+        kwargs['team'] = team
+        context.update(kwargs)
+        return super().get_context_data(**context)
```
fjsj: Here you should just `return context`

backend/battle/views.py
```diff
@@ -69,3 +69,20 @@ def get_queryset(self):
...
+    def get_context_data(self, **kwargs):
+        context = {}
+        if self.object:
+            context['object'] = self.object
+            context_object_name = self.get_context_object_name(self.object)
+            if context_object_name:
+                context[context_object_name] = self.object
+        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
+        kwargs['team'] = team
+        context.update(kwargs)
```
fjsj: I think you can change this to `context['team'] = team`

fjsj (changes requested):
Please check comments

backend/battle/views.py
DeborahAlmeida: Thank you!

backend/battle/views.py
DeborahAlmeida: It's really misleading, thank you :)

rsarai (approved):

PR #10, opened 2021-06-17
create an email var

DeborahAlmeida:
## Objective and Motivation
- Create an email var to manage the sending of emails


backend/pokebattle/settings/production.py
```diff
@@ -30,6 +30,8 @@
 EMAIL_HOST_PASSWORD = config("SENDGRID_PASSWORD")
 EMAIL_PORT = 587
 EMAIL_USE_TLS = True
+EMAIL = config("EMAIL")
```
fjsj: I would change to `FROM_EMAIL`

fjsj (changes requested):
Please check comments

backend/pokebattle/settings/production.py
DeborahAlmeida: Thank you!

rsarai (approved):

PR #11, opened 2021-06-18
Feature: Integration with Pokemon API

DeborahAlmeida:
## Objective and Motivation
- Integrate with pokemon API
- Verify if the pokemon does not exist in the system database
- Insert on the database the new pokemon

### DIFF
--- a/backend/battle/battles/battle.py
+++ b/backend/battle/battles/battle.py
@@ -80,3 +80,8 @@ def set_winner(winner, battle):
     battle.winner = winner
     battle.save()
     result_battle(battle)
+
+
+def get_pokemon_object(pokemon_id):
+    pokemon_object = Pokemon.objects.get(pokemon_id=pokemon_id)
+    return pokemon_object
--- a/backend/battle/battles/battle.py
+++ b/backend/battle/battles/battle.py
@@ -38,11 +38,10 @@ def validate_sum_pokemons(pokemons):
 
 
 def verify_pokemon_is_saved(pokemons):
-    pokemons_not_saved = []
     for pokemon in pokemons:
         on_database = verify_pokemon_exists_on_database(pokemon)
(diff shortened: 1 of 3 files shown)


backend/battle/battles/battle.py
```diff
@@ -29,14 +31,57 @@ def get_winner_for(team_creator, team_opponent):
...
+        else:
+            pokemons_saved.append(on_database)
+    for pokemon in pokemons_saved:
+        pokemon_point = sum_points_pokemon_on_database(pokemon)
         total_points += pokemon_point
+    if total_api_pokemon_point:
+        total_points += total_api_pokemon_point
+    if total_points <= 600:
+        for pokemon in pokemons_not_saved:
+            create_pokemon_on_database(pokemon)
```
fjsj, line 44: This logic is a bit confusing. You're at the same time validating and saving?
Consider separating those things, like:
1. First save what is not saved (separate this on another function)
2. Then have a single logic to sum those points

fjsj (changes requested):
Please check the comments

backend/battle/battles/battle.py
DeborahAlmeida: Well, I need to sum those points before saving the pokemon from API, because if the points sum more than 600, I cannot save the pokemon selected

DeborahAlmeida: Now I understood your argument. After your advice, I think I fixed this misleading

DeborahAlmeida: Thank you!

fjsj: @rsarai could you please review the new logic?
Code-wise is fine, but I'm not sure if the behavior needs to be like this.

rsarai (approved):

PR #16, opened 2021-06-25
send email to opponent when create a battle

DeborahAlmeida:
## Objective and Motivation
- As a selected opponent I receive an email when I'm invited to a match


backend/battle/battles/email.py
```diff
@@ -11,3 +11,14 @@ def result_battle(battle):
             'winner': battle.winner,
         },
     )
+
+
+def send_invite_email(opponent, creator):
+    send_templated_mail(
+        template_name='invite_challenge',
+        from_email=settings.FROM_EMAIL,
+        recipient_list=[opponent],
```
fjsj: Should be `opponent.email` here.
This is only working right now because your `__str__` is `self.email`. You shouldn't rely on that.

fjsj (changes requested):
Please check comments

backend/battle/battles/email.py
DeborahAlmeida: Oh, you told me once and I forgot to apply, thanks to remember me!

fjsj (approved):
LGTM, thanks

PR #17, opened 2021-06-28
Feature/signup

DeborahAlmeida:
## Objective and Motivation
- The user can signup to the app


backend/battle/templates/battle/base.html
```diff
@@ -196,14 +199,20 @@
             <p class="p_user">{{user}}</p>
         </a>
     </div>
+    {% else %}
+    <div class="user">
+        <a class ="anchor_user" href="{% url 'signup' %}">
+            <img class="img_user" style="width:50px; height:50px" src="https://www.greatmining.com/images/signup.png">
```
fjsj: Please avoid inline styles.
For images, it's fine to add width and height directly as attributes.

backend/battle/templates/battle/user/signup_form.html
```diff
@@ -0,0 +1,227 @@
+{% load static %}
+<!DOCTYPE html>
```
fjsj: Any reasoning for re-declaring the whole html again instead of using a base html file and extends?

Note your base.html probably shouldn't be inside `battle/base.html` as well.

fjsj (changes requested):
Please check comments

backend/battle/templates/battle/base.html
DeborahAlmeida: thanks!

backend/battle/templates/battle/user/signup_form.html
DeborahAlmeida: Nop, it was wrong, thanks :)

backend/battle/templates/battle/base.html
```diff
@@ -196,14 +16,20 @@
...
```
fjsj: There are still inline stiles here and in other places.

fjsj (changes requested):
Please check comments

backend/battle/templates/battle/user/signup_form.html
fjsj: This is also unsolved

fjsj (approved):
LGTM

PR #18, opened 2021-07-01
Feature/img url

DeborahAlmeida:
## Objective and Motivation
- Add img_url column to pokemon's model
- Add pokemon's info to template of detail battle


backend/battle/views.py
```diff
@@ -74,8 +75,12 @@ class BattleDetail(DetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
-        context['team'] = team
+        team_user = Team.objects.filter(battle=self.object, trainer=self.request.user)
```
fjsj: Will the user always be the creator or opponent? If so, you're making an unncessary query here.
You can instead move this two lines below and do:
```python
team_user = team_creator if self.request.user == self.object.creator else team_opponent
```

backend/battle/battles/base_stats.py
```diff
@@ -33,3 +34,18 @@ def sum_all_pokemons(pokemons):
 def sumValid(pokemon):
     sum_result = pokemon["attack"] + pokemon["defense"] + pokemon["hp"]
     return sum_result
+
+
+def verify_team_exists(battle, trainer):
```
fjsj: Could you please clarify what you're doing here? The name isn't very readable to me. This doesn't simply verifies, but also returns the pokemon team as a dict.

backend/battle/battles/base_stats.py
```diff
@@ -33,3 +34,18 @@ def sum_all_pokemons(pokemons):
...
+    pokemons_team = {
+        "pokemon_1": 0,
+        "pokemon_2": 0,
+        "pokemon_3": 0,
+    }
+    team = Team.objects.filter(battle=battle, trainer=trainer)
```
fjsj: Use `Team.objects.prefetch_related('pokemons').filter(battle=battle, trainer=trainer)` to avoid an additional query for each Pokemon.

fjsj (changes requested):
Please check comments

backend/pokemon/models.py
```diff
@@ -8,3 +8,4 @@ class Pokemon(models.Model):
     attack = models.IntegerField()
     defense = models.IntegerField()
     hp = models.IntegerField()
+    img_url = models.CharField(max_length=100, blank=True, )
```
fjsj: max_length=100 is quite low for an URL. I suggest changing to 255

backend/battle/views.py
DeborahAlmeida: This way is easier to understand, thanks!

DeborahAlmeida: I tried this right now and I realized the team_creator and team_opponent are not a user object, they are an object with pokemons related, so if I do this logic, the code breaks

backend/battle/battles/base_stats.py
DeborahAlmeida: Sorry, I'm going to change this function name. This function get all pokemons related to a team, if the team doesn't exist, it will be returned a default value to a dict.

backend/battle/battles/base_stats.py
DeborahAlmeida: Thanks, the prefetch_related and select_related' approach is still a bit confusing to me

backend/pokemon/models.py
DeborahAlmeida: okay

backend/battle/views.py
fjsj: The only thing you do with `team_user` is checking if it exists in the template.
But you already check if the team exists inside `get_pokemons_team`. So you're repeating a query without need.

**I think if you check if self.request.user is either the creator or opponent, and also check if one of pokemons_creator and pokemons_opponent is empty, you get the same result.**

So you can switch `team_user` to `user_has_team` and do: `user_has_team = self.request.user in [self.object.creator, self.object.opponent] and (not pokemons_creator or not pokemons_opponent)`

Please double check and lmk if you have questions.

backend/battle/battles/base_stats.py
fjsj: No worries, check http://pythonclub.com.br/django-introducao-queries.html

fjsj (changes requested):
Please check comments

backend/battle/views.py
DeborahAlmeida: Thanks, I put this code and I tested it, it didn't work well, because for example, when the -- user logged -- is the creator and the user has not a team yet, but the opponent does,  "user_has_team" turns to True, instead of False. Therefore I change the old logic and deleted the unnecessary query. Let me know if you think it's clear to understand the logic. Thanks again!

backend/battle/battles/base_stats.py
DeborahAlmeida: thanks :)

backend/battle/battles/base_stats.py
```diff
@@ -33,3 +34,20 @@ def sum_all_pokemons(pokemons):
 def sumValid(pokemon):
     sum_result = pokemon["attack"] + pokemon["defense"] + pokemon["hp"]
     return sum_result
+
+
+def get_pokemons_team(battle, trainer):
+    pokemons_team = {
+        "pokemon_1": 0,
+        "pokemon_2": 0,
+        "pokemon_3": 0,
```
fjsj: It is strange the default is `0` here. Should be `None`.

backend/battle/templates/battle/battle_detail.html
```diff
@@ -9,11 +9,139 @@ <h2 class="subtitle_detail">Opponent: {{battle.opponent}}</h2>
       <img width=60 height=60 class="img_detail" src="https://image.flaticon.com/icons/png/512/2119/2119019.png" alt="trofeu">
       <h2 class="subtitle_detail">The winner is {{battle.winner}}</h2> 
   {% endif %}
-  {% if not team %}
+  {% if pokemons_user == False %}
```
fjsj: `if not pokemons_user` is probably better

backend/battle/views.py
```diff
@@ -74,8 +74,17 @@ class BattleDetail(DetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        team = Team.objects.filter(battle=self.object, trainer=self.request.user)
-        context['team'] = team
+
+        pokemons_creator = get_pokemons_team(self.object, self.object.creator)
+        pokemons_opponent = get_pokemons_team(self.object, self.object.opponent)
+        pokemons_user = 0
```
fjsj: Use `None` instead

fjsj (changes requested):
Please check comments

backend/battle/views.py
DeborahAlmeida: thank you :)

fjsj (approved):
LGTM

PR #20, opened 2021-07-05
Feature/autocomplete pokemon

DeborahAlmeida:
##Objective and Motivation
- Add autocomplete to pokemon form
- Feature of type ahead


backend/battle/forms.py
```diff
@@ -1,11 +1,19 @@
 from django import forms
+import requests
+from urllib.parse import urljoin
+import requests
```
fjsj: There are repeated `import requests` here

backend/battle/forms.py
```diff
@@ -36,41 +44,35 @@ class Meta:
...
-        min_value=1,
-        max_value=898,
-    )
-
-    def __init__(self, *args, **kwargs):
-        super(TeamForm, self).__init__(*args, **kwargs)
-        self.fields['battle'].widget = forms.HiddenInput()
-        self.fields['trainer'].widget = forms.HiddenInput()
+    battle = forms.ModelChoiceField(queryset=Battle.objects.all())
+    trainer = forms.ModelChoiceField(queryset=User.objects.all())
```
fjsj: Those used to be hidden, any reason for not being now?

backend/battle/forms.py
```diff
@@ -36,41 +44,35 @@ class Meta:
...
+        pokemon_2 = self.cleaned_data.get('pokemon_2')
+        pokemon_3 = self.cleaned_data.get('pokemon_3')
+
+        if not pokemon_1 or not pokemon_2 or not pokemon_3:
+            raise forms.ValidationError('ERROR: Select all pokemons')
+
+        for pokemon in [pokemon_1, pokemon_2, pokemon_3]:
+            response = requests.get(urljoin(settings.POKE_API_URL, pokemon))
+            if response.status_code == 404:
+                raise forms.ValidationError('ERROR: Type the correct pokemon name ')
```
fjsj: This `for` and this response only checks if the pokemon exist in the API, right?
So I think this should be moved to a service function and you should call the service here.

The service can return `False` if any pokemon doesn't exist, and in that case you can raise this exception.

backend/battle/forms.py
```diff
@@ -36,41 +44,35 @@ class Meta:
...
             [
                 cleaned_data['pokemon_1'],
                 cleaned_data['pokemon_2'],
                 cleaned_data['pokemon_3']
             ]
         )
-        obj_battle = cleaned_data['battle']
-        if cleaned_data['trainer'] not in (obj_battle.creator, obj_battle.opponent):
+
+        if cleaned_data['trainer'] not in (cleaned_data['battle'].creator, cleaned_data['battle'].opponent):
```
fjsj: This code was more readable before.

backend/battle/forms.py
```diff
@@ -83,15 +83,14 @@ def clean(self):
...
-        cleaned_data['pokemon_3_object'] = get_pokemon_object(cleaned_data['pokemon_3'])
+        cleaned_data['pokemon_1_object'] = Pokemon.objects.get(name=cleaned_data['pokemon_1'])
+        cleaned_data['pokemon_2_object'] = Pokemon.objects.get(name=cleaned_data['pokemon_2'])
+        cleaned_data['pokemon_3_object'] = Pokemon.objects.get(name=cleaned_data['pokemon_3'])
         return cleaned_data
 
     def save(self, commit=True):
         data = self.clean()
-        instance = super().save()
+        instance = Team.objects.create(battle=data['battle'], trainer=data['trainer'])
```
fjsj: Why can't you call super anymore?

backend/battle/templates/battle/pokemon_form.html
```diff
@@ -12,12 +13,46 @@
...
             <form method="POST" class="battle-form battle-form-pokemon">{% csrf_token %}
-                {{ form.as_p }}
-                            
+                {{ form.non_field_errors }}
+                <input hidden name="battle" id="battle" value="{{battle}}">
+                <input hidden name="trainer" id="trainer" value="{{trainer}}">
+                <p>
+                    <label for="id_pokemon_1">Pokemon 1:</label>
+                    <input id="pokemon_1" name="pokemon_1" type="text" list="pokemons">
+                    <datalist id="pokemons" name="pokemon_1">
```
fjsj, line 23: This was needed for the autocomplete?

backend/battle/forms.py
DeborahAlmeida: Great, I'll make this improvement.

backend/battle/views.py
```diff
@@ -43,11 +45,13 @@ class TeamView(LoginRequiredMixin, CreateView):
...
-    def get_initial(self):
-        obj_battle = get_object_or_404(Battle, pk=self.kwargs['pk'])
-        obj_trainer = self.request.user
-        self.initial = {"battle": obj_battle, "trainer": obj_trainer}
-        return self.initial
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        pokemons = Pokemon.objects.all()
+        context['pokemons'] = pokemons
+        context['battle'] = self.kwargs['pk']
```
fjsj: `battle_pk` is a better name for the context variable. `battle` implies an object.

backend/battle/views.py
```diff
@@ -43,11 +45,13 @@ class TeamView(LoginRequiredMixin, CreateView):
...
+        context['trainer'] = self.request.user.id
```
fjsj: Same here.

fjsj (changes requested):
Please check comments.

backend/battle/forms.py
DeborahAlmeida: I deleted the obj_battle because now I brought directly from view. Do you think it's still better to store the cleaned_data['battle'] in obj_battle?

backend/battle/forms.py
DeborahAlmeida: thanks!

backend/battle/forms.py
DeborahAlmeida: I put this attribute directly on the template form.

backend/battle/forms.py
DeborahAlmeida: You're right, I changed this now, thanks!

backend/battle/templates/battle/pokemon_form.html
DeborahAlmeida: Yes

backend/battle/forms.py
fjsj: I see what you want to achieve. In this case, to simplify the code, you can use the hidden widget like this:
```python
battle = forms.ModelChoiceField(widget=forms.HiddenInput(), queryset=Battle.objects.all())
trainer = forms.ModelChoiceField(widget=forms.HiddenInput(), queryset=User.objects.all())
```

And in template, there's no need for manually using `<input>` for the hidden ones. Use instead this:
```
{% for hidden in form.hidden_fields %}
    {{ hidden }}
{% endfor %}
```

Note now you don't need anymore battle and trainer on template.

backend/battle/forms.py
fjsj: Yes, because it's just an additional variable.

fjsj: I mean just doing `obj_battle = cleaned_data['battle']`

backend/battle/forms.py
DeborahAlmeida: Thank you!

backend/battle/forms.py
DeborahAlmeida: great!

backend/battle/views.py
DeborahAlmeida: Sure, thank you again :)

backend/battle/forms.py
DeborahAlmeida: Hey, when I deleted the tags hidden on the template, I got a KeyError. It seems I did not have the battle and trainer values. Do you know what this can be?

backend/battle/views.py
DeborahAlmeida: thanks :)

backend/battle/forms.py
fjsj: Did you put the `{% for hidden in form.hidden_fields %}` I said above? Note this goes inside the <form>

fjsj: You should put this AND remove the hidden tags.

DeborahAlmeida: Yeah, I'm going to try again to make sure.

backend/pokemon/helpers.py
```diff
@@ -13,5 +13,14 @@ def get_pokemon_from_api(poke_id):
         "hp": data["stats"][0]["base_stat"],
         "name": data["name"],
         "img_url": data["sprites"]["front_default"],
+        "pokemon_id": data["id"]
     }
     return info
+
+
+def verify_pokemon_exists_api(pokemons):
```
fjsj: those pokemons are pokemon names actually, right? worth making this clearer. 
consider change the variable name to `poke_names`

fjsj (changes requested):
Just a small comment, overall looking great

backend/pokemon/helpers.py
DeborahAlmeida: Great! I'm going to change this :D

fjsj (approved):
LGTM

PR #21, opened 2021-07-07
Feature/position pokemon

DeborahAlmeida:
## Objective and Motivation
- The user can choose the pokemon's position on battle


backend/battle/battles/base_stats.py
```diff
@@ -44,10 +44,10 @@ def get_pokemons_team(battle, trainer):
     }
     team = Team.objects.prefetch_related('pokemons').filter(battle=battle, trainer=trainer)
     if team:
-        pokemons_team_query = team[0].pokemons.all()
-        pokemons_team["pokemon_1"] = pokemons_team_query[0]
-        pokemons_team["pokemon_2"] = pokemons_team_query[1]
-        pokemons_team["pokemon_3"] = pokemons_team_query[2]
+        pokemon_team = team[0].teams.all().order_by('order').prefetch_related('pokemon')
```
fjsj: This is a bit strange because you're not using the `Team` instance anymore.
I think you can query instead directly from `PokemonTeam` in line 45. Please check if you can make this refactor.

backend/battle/battles/battle.py
```diff
@@ -12,13 +12,19 @@ def run_battle(battle):
 
 
 def get_winner_for(team_creator, team_opponent):
-    creator_pokemons = team_creator.pokemons.all()
-    opponent_pokemons = team_opponent.pokemons.all()
+
+    pkn_team_cre = team_creator.teams.all().order_by('order').prefetch_related('pokemon')
```
fjsj: Those abbreviations are a bit unreadable IMO. It's better to use names like `creator_pkn_team_qs` for example.

backend/battle/forms.py
```diff
@@ -90,11 +109,11 @@ def save(self, commit=True):
         data = self.clean()
```
fjsj: `data = self.clean()` -> `cleaned_data = self.clean()`

Because forms have both data and cleaned_data, so this will avoid confusion.

backend/battle/templates/battle/pokemon_form.html
```diff
@@ -17,33 +17,57 @@
...
-                    <datalist id="pokemons" name="pokemon_3">
-                        {% for results in pokemons %}
-                            <option value="{{results.name}}">
-                        {% endfor %}
-                    </datalist>
-                </p>
+                <div class="form_pokemon_position">
+                    <p>
+                        <label for="id_pokemon_1">Pokemon 1:</label>
+                        <input id="pokemon_1" name="pokemon_1" type="text" list="pokemons">
```
fjsj, line 24: No form errors for this input here.

backend/requirements.txt
```diff
@@ -107,4 +107,4 @@ wcwidth==0.2.5
 whitenoise==5.2.0
 wrapt==1.12.1
 xlrd==2.0.1
-xlwt==1.3.0
+xlwt==1.3.0
```
fjsj: I don't see the iteration_utilities in those requirements.
Also, why there's requirements.txt at root and this other one? I think there should be only one at root.

fjsj (changes requested):
Please check comments.

backend/battle/battles/base_stats.py
DeborahAlmeida: I need to use the Team because in line 45 I verify if a team exists, if not, it'll return false.
If a team exists, I get the data of pokemon_team filtering by the team

backend/battle/battles/battle.py
DeborahAlmeida: I put the abbreviations because pylint was giving many characters error, so I need to shorten the variable names as much as possible

backend/requirements.txt
DeborahAlmeida: I was put on the wrong path. I will fix it, thanks :)

backend/battle/battles/base_stats.py
fjsj: But if the PokemonTeam for that user exists, then the Team exists.

So consider if you can filter by PokemonTeam crossing Team, like this: `PokemonTeam.objects.filter(team__battle=battle, team__trainer=trainer).select_related('pokemon')`

backend/battle/battles/battle.py
fjsj: That's a bad pylint error. You can reconfigure it to avoid it. LMK if you have trouble.

backend/battle/battles/base_stats.py
DeborahAlmeida: It works, thanks!

backend/battle/battles/battle.py
DeborahAlmeida: Yes, this used to complicate my development, but I remember Xima commented this limit of character is a good practice, should I reconfigure the limit to more or change my variable names?

backend/battle/templates/battle/pokemon_form.html
DeborahAlmeida: I defined validation error for pokemons typed wrong way and file empty. So, I don't understand, what kind of form error?

backend/battle/forms.py
DeborahAlmeida: I don't understand, should I use data instead of cleaned_data?

backend/battle/battles/battle.py
fjsj: cool, configure to more, like 25

backend/battle/forms.py
fjsj: just the variable name, call it cleaned_data

backend/battle/templates/battle/pokemon_form.html
fjsj: I mean printing the form errors like `{{ form.errors.pokemon_1 }}`

backend/battle/battles/battle.py
DeborahAlmeida: thanks :)

requirements.txt
```diff
@@ -4,19 +4,19 @@
 #
```
fjsj, line 4: Still no iteration_utilities in this file. Double-check how to add a requirement: https://github.com/DeborahAlmeida/pokebattle#adding-new-pypi-libs

DeborahAlmeida: I'm not using iteration_utilities anymore. But I'm going to check it, thank you!

backend/battle/templates/battle/pokemon_form.html
DeborahAlmeida: Alright, thanks :)

backend/battle/forms.py
DeborahAlmeida: Sure, I'll change this.

requirements.txt
DeborahAlmeida: Thank you, I realized where I was doing wrong.

fjsj (approved):
LGTM, but with concerns discussed via Slack

PR #22, opened 2021-07-07
Feature/send invite opponent

DeborahAlmeida:
## Objective and Motivation
- As a Trainer I can invite an opponent by email if they are not yet in the system


backend/battle/forms.py
```diff
@@ -53,6 +58,10 @@ def clean(self):
         pokemon_3 = self.cleaned_data.get('pokemon_3')
         obj_battle = cleaned_data['battle']
         obj_trainer = cleaned_data['trainer']
+        trainer_on_database = User.objects.filter(email=obj_trainer)
```
fjsj, line 61: Why do you need those lines? `obj_trainer` should already be a reference to the trainer in the DB.

backend/battle/battles/email.py
```diff
@@ -14,10 +15,16 @@ def result_battle(battle):
 
 
 def send_invite_email(opponent, creator):
+    user_on_database = User.objects.filter(email=opponent)
+    opponent_email = None
+    if not user_on_database:
+        opponent_email = opponent
+    else:
+        opponent_email = user_on_database[0].email
```
fjsj, line 23: Why do you have to add those lines?

fjsj:
I'm checking with Rob the proper way to implement this.

PR #23, opened 2021-07-08
send invite to user

DeborahAlmeida:
## Objective and Motivation
- As a Trainer I can invite an opponent by email if they are not yet in the system


backend/battle/forms.py
```diff
@@ -20,13 +27,43 @@ def __init__(self, *args, **kwargs):
         super(BattleForm, self).__init__(*args, **kwargs)
         self.fields['creator'].widget = forms.HiddenInput()
 
+    def clean_opponent(self):
+        opponent_email = self.cleaned_data["opponent"]
+        try:
+            opponent = User.objects.get(email=opponent_email)
+        except User.DoesNotExist:
+            self.is_guest = True
```
fjsj: I'm not sure if this is completely safe to do.
Instead try the following: set `opponent.is_guest = True` and read new custom attr this below on `clean`.

fjsj: No need for a migration, this is just a "temporary" attr.

backend/battle/forms.py
```diff
@@ -20,13 +27,43 @@ def __init__(self, *args, **kwargs):
...
         if 'creator' not in cleaned_data:
             raise forms.ValidationError("ERROR: You need to be logged.")
 
         if cleaned_data['opponent'] == cleaned_data['creator']:
             raise forms.ValidationError("ERROR: You can't challenge yourself.")
+        return cleaned_data
+
+    def save(self, commit=True):
+        cleaned_data = self.clean()
+        instance = super().save()
```
fjsj, line 55: Just a nitpick: you're not passing to super the commit param. Not a problem in this case here, but beware this in the future.

backend/battle/forms.py
```diff
@@ -20,13 +27,43 @@ def __init__(self, *args, **kwargs):
...
+        opponent = cleaned_data["opponent"]
+        if self.is_guest:
+            invite_form = PasswordResetForm(data={"email": opponent.email})
+            invite_form.is_valid()
+            invite_form.save(
+                self, subject_template_name='registration/password_reset_subject.txt',
```
fjsj: You should use your own templates for subject and email. And they should have different file names and contents.
Make the contents reflect the behavior here, which is an invite, not a password reset.

backend/battle/forms.py
```diff
@@ -20,13 +27,43 @@ def __init__(self, *args, **kwargs):
...
+                email_template_name='registration/password_reset_email.html',
+                use_https=False, token_generator=default_token_generator,
+                from_email='deborah.mendonca@vinta.com.br',
```
fjsj: This from email here is hardcoded. Maybe use instead the one you have on settings?

backend/battle/forms.py
DeborahAlmeida: But the user can be a guest or not, how can I verify this on clean?

backend/battle/forms.py
```diff
@@ -20,13 +27,43 @@ def __init__(self, *args, **kwargs):
...
+        cleaned_data = self.clean()
+        instance = super().save()
+        opponent = cleaned_data["opponent"]
+        if self.is_guest:
+            invite_form = PasswordResetForm(data={"email": opponent.email})
+            invite_form.is_valid()
+            invite_form.save(
+                self, subject_template_name='registration/password_reset_subject.txt',
+                email_template_name='registration/password_reset_email.html',
+                use_https=False, token_generator=default_token_generator,
```
fjsj: Those two parameters need not to be set, as they have default values already. You can just remove this line I guess.

backend/battle/forms.py
DeborahAlmeida: Got it, thanks!

backend/battle/urls.py
```diff
@@ -20,4 +21,9 @@
     path("battle/detail/<int:pk>/", BattleDetail.as_view(), name="battle_detail"),
     path('signup/', BattleSignUp.as_view(), name='signup'),
     path('signup/sucess', SignUpSucess.as_view(), name='signup_sucess'),
+    path(
+        'reset/<uidb64>/<token>/',
+        auth_views.PasswordResetConfirmView.as_view(
+            success_url=reverse_lazy(
+                'password_reset_complete')), name='password_reset_confirm'),
```
fjsj: Is this working? I don't see the password_reset_complete view/url here.

fjsj (changes requested):
Please check comments

backend/battle/forms.py
fjsj: set `opponent.is_guest = False` above after `opponent = User.objects.get(email=opponent_email)`.

backend/battle/forms.py
DeborahAlmeida: Yes, I'll do it, thanks :)

backend/battle/forms.py
DeborahAlmeida: Thanks :) I'll delete it

backend/battle/forms.py
DeborahAlmeida: Got it :)

backend/battle/urls.py
DeborahAlmeida: It works

backend/battle/forms.py
DeborahAlmeida: Okay :)

backend/battle/templates/battle/battle_form.html
```diff
@@ -13,8 +13,19 @@
...
-                        
+            {{ form.non_field_errors }}
+            {% for hidden in form.hidden_fields %}
+                {{ hidden }}
+            {% endfor %}
+            <p>
+                <label>Opponent:</label>
+                <input id="opponent" name="opponent" type="text" list="users">
+                <datalist id="users" name="opponent">
+                    {% for results in users %}
```
fjsj: `for user` instead of `for results` is best here, because the plural implies multiple users.

backend/battle/urls.py
fjsj: but which page the users sees when password reset is confirmed?

fjsj (changes requested):
Looking good overall. Just a couple of fixes missing, please check comments.

backend/battle/urls.py
fjsj: When you do that, the user will think they're resetting the password, when in fact they are setting it for the first time, so they'll be confused. You must redefine and have both behaviors: password reset and password set when first invited.

You can get the URLs and views from https://github.com/django/django/blob/1f4908b01aaf7b209521d40ca0f3eb48f2ef1988/django/contrib/auth/urls.py and re-implement as needed.

fjsj:
Conflicts here.

backend/battle/urls.py
DeborahAlmeida: Thank you, I'll make this improvement

backend/battle/templates/battle/battle_form.html
DeborahAlmeida: You're right, I'm going to fix this :)

backend/battle/views.py
```diff
@@ -106,3 +128,119 @@ class BattleSignUp(CreateView):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update({
+            'title': self.title,
+            **(self.extra_context or {})
+        })
+        return context
+
+
+class PasswordCreateConfirmView(PasswordContextMixin, FormView):
```
fjsj: You don't need to copy the whole code. You can make a new class, inherit from the existing one, and change only what you need.

fjsj (dismissed):
Looking good overall, but I suggest using inheritance to avoid copying a lot of code.

backend/battle/views.py
DeborahAlmeida: Thank you!

rsarai (approved):

PR #26, opened 2021-07-20
create test to sum point of pokemons

DeborahAlmeida:
## Objective and Motivation
- Create a test to verify if the points of the chosen pokemons sum greater than 600

### DIFF
--- a/backend/battle/tests.py
+++ b/backend/battle/tests.py
@@ -1,3 +1,14 @@
-# from django.test import TestCase
+from django.test import TestCase
+from battle.battles.battle import validate_sum_pokemons
 
-# Create your tests here.
+
+class PokemonsSumTest(TestCase):
+    def test_sum(self):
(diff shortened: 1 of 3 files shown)


backend/battle/tests.py
```diff
@@ -1,3 +1,14 @@
-# from django.test import TestCase
+from django.test import TestCase
+from battle.battles.battle import validate_sum_pokemons
 
-# Create your tests here.
+
+class PokemonsSumTest(TestCase):
+    def test_sum(self):
+        pokemons = ['bulbasaur', 'pikachu', 'pidgeot']
+        sum_pokemons = validate_sum_pokemons(pokemons)
```
rsarai, line 8: Let's not test getting the pokemon from the API. It's not a good practice to allow tests to make requests for external APIs. Please get the pokemon's from the database

DeborahAlmeida: Thanks :)

PR #27, opened 2021-07-21
test get winner

DeborahAlmeida:
## Objective and Motivation
- Test to verify if it returns the battle's right winner


backend/battle/tests.py
```diff
@@ -1,3 +1,29 @@
-# from django.test import TestCase
 
-# Create your tests here.
+from django.test import TestCase, Client
```
rsarai: Please create a folder for tests and name this according to the module being tested

backend/battle/tests.py
```diff
@@ -1,3 +1,29 @@
...
+        self.team_creator = baker.make('battle.Team', battle=self.battle, trainer=self.creator)
+        self.team_opponent = baker.make('battle.Team', battle=self.battle, trainer=self.opponent)
+        PokemonTeam.objects.create(team=self.team_creator, pokemon=self.pokemon_1, order=1)
+        PokemonTeam.objects.create(team=self.team_creator, pokemon=self.pokemon_2, order=2)
+        PokemonTeam.objects.create(team=self.team_creator, pokemon=self.pokemon_3, order=3)
+        PokemonTeam.objects.create(team=self.team_opponent, pokemon=self.pokemon_1, order=1)
+        PokemonTeam.objects.create(team=self.team_opponent, pokemon=self.pokemon_2, order=2)
+        PokemonTeam.objects.create(team=self.team_opponent, pokemon=self.pokemon_3, order=3)
+
+    def test_verify_if_returns_correct_winner(self):
```
rsarai: The happy path is not enough, we should test for different cases as well. An example would be: teams with the same amount of points, teams with different points, different orders, and error scenarios.

backend/battle/tests.py
DeborahAlmeida: Got it :)

backend/battle/tests.py
DeborahAlmeida: Great! Are there some good practices and rules to make enough scenarios for tests?

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+        self.team_opponent = baker.make('battle.Team', battle=battle, trainer=self.opponent)
+        self.pokemon_1, self.pokemon_2, self.pokemon_3 = baker.make(
+            "pokemon.Pokemon",
+            attack=60, defense=45, hp=50, _quantity=3)
+
+    def test_verify_if_returns_creator_as_winner_when_draw(self):
+        pokemons = [self.pokemon_1, self.pokemon_2, self.pokemon_3]
+        count_creator = 0
+        count_opponent = 0
+        while count_creator < 3:
```
rsarai: I'm not particularly eager to create test info inside while loops, for situations where I really need to I use a for. You can use a for with enumerate

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+            PokemonTeam.objects.create(
+                team=self.team_creator,
+                pokemon=pokemons[count_creator],
+                order=count_creator)
+            count_creator = count_creator + 1
+
+        while count_opponent < 3:
```
rsarai: same as before

backend/battle/tests/test_service.py
rsarai: The reason is that it makes the intention more clear

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+            PokemonTeam.objects.create(
+                team=self.team_opponent,
+                pokemon=pokemons[count_opponent],
+                order=count_opponent)
+            count_opponent = count_opponent + 1
+
+        winner = get_winner_for(self.team_creator, self.team_opponent)
+        self.assertEqual(winner, self.team_creator)
+
+    def test_verify_if_not_returns_winner_when_different_points(self):
```
rsarai: shouldn't this be: `test_verify_if_not_returns_creator_as_winner_when_different_points`?

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+        pokemon_4, pokemon_5, pokemon_6 = baker.make(
+            "pokemon.Pokemon",
+            attack=100, defense=35, hp=30, _quantity=3)
+        pokemons_creator = [self.pokemon_1, self.pokemon_2, self.pokemon_3]
+        pokemons_opponent = [pokemon_4, pokemon_5, pokemon_6]
+        count_creator = 0
+        count_opponent = 0
+        while count_creator < 3:
```
rsarai: Same as before

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+            PokemonTeam.objects.create(
+                team=self.team_creator,
+                pokemon=pokemons_creator[count_creator],
+                order=count_creator)
+            count_creator = count_creator + 1
+
+        while count_opponent < 3:
```
rsarai: Same as before

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+            count_creator = count_creator + 1
+
+        while count_opponent < 2:
+            PokemonTeam.objects.create(
+                team=self.team_opponent,
+                pokemon=pokemons_opponent[count_opponent],
+                order=count_opponent)
+            count_opponent = count_opponent + 1
+        try:
+            get_winner_for(self.team_creator, self.team_opponent)
```
rsarai: This is not actually testing anything since the test would not fail if the exception is not thrown. You can use the `assertRaises` here.

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
+        pokemon_4, pokemon_5, pokemon_6 = baker.make(
+            "pokemon.Pokemon",
+            attack=100, defense=35, hp=30, _quantity=3)
+        pokemons_creator = [self.pokemon_1, self.pokemon_2, self.pokemon_3]
+        pokemons_opponent = [pokemon_4, pokemon_5, pokemon_6]
+        count_creator = 0
+        count_opponent = 0
+        while count_creator < 3:
```
rsarai: same as before

backend/battle/tests/test_service.py
```diff
@@ -0,0 +1,88 @@
...
+            PokemonTeam.objects.create(
+                team=self.team_creator,
+                pokemon=pokemons_creator[count_creator],
+                order=count_creator)
+            count_creator = count_creator + 1
+
+        while count_opponent < 2:
```
rsarai: same as before

rsarai (changes requested):
Great improvement. Left some more comments

backend/battle/tests/test_service.py
DeborahAlmeida: It's wrong, I'm going to fix it. It should be `test_verify_if_returns_creator_as_winner_when_different_points`

backend/battle/tests/test_service.py
DeborahAlmeida: Thank you for this suggestion!

rsarai (approved):

PR #28, opened 2021-07-21
Feature/test create battle view

DeborahAlmeida:
## Objective and Motivation
- Test if it creates a battle correctly

### DIFF
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -24,6 +24,7 @@ class BattleForm(forms.ModelForm):
 
     def clean(self):
         cleaned_data = super().clean()
+        import ipdb; ipdb.set_trace()
         if 'creator' not in cleaned_data:
             raise forms.ValidationError("ERROR: You need to be logged.")
 
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -24,7 +24,6 @@ class BattleForm(forms.ModelForm):
 
     def clean(self):
         cleaned_data = super().clean()
-        import ipdb; ipdb.set_trace()
         if 'creator' not in cleaned_data:
             raise forms.ValidationError("ERROR: You need to be logged.")
 
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -46,6 +46,9 @@ class BattleForm(forms.ModelForm):
         if 'creator' not in cleaned_data:
             raise forms.ValidationError("ERROR: You need to be logged.")
 
+        if 'opponent' not in cleaned_data:
+            raise forms.ValidationError("ERROR: You need to choose an opponent")
+
         if cleaned_data['opponent'] == cleaned_data['creator']:
             raise forms.ValidationError("ERROR: You can't challenge yourself.")
         return cleaned_data
--- a/backend/battle/tests.py
+++ b/backend/battle/tests.py
@@ -1,3 +1,48 @@
-# from django.test import TestCase
+from django.test import TestCase, Client
+from model_bakery import baker
+from users.models import User
+from battle.models import Battle
 
-# Create your tests here.
+'''
+class ListBattlesTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+        self.battle = baker.make('battle.Battle')
+        self.user = User.objects.create(email='deborah.mendonca@vinta.com.br', password='admin')
+        self.user.set_password('admin')
+        self.user.save()
+    
+    def test_login_user_can_acess_battle_list(self):
+        # import ipdb; ipdb.set_trace()
+        self.client.login(username=self.user.email, password='admin')
+        response = self.client.get('/battle/list/')
+        self.assertEqual(response.status_code, 200)
+    
+    def test_data_returns_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=2)
+        response = self.client.get('/battle/list/')
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
+'''
+
+class BattleCreateViewTest(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.client = Client()
+        self.creator = User.objects.create(email='deborah.mendonca@vinta.com.br', password='admin')
+        self.creator.set_password('admin')
+        self.creator.save()
+        self.opponent = baker.make('users.User')
+    
+    def test_create_battle_successfully(self):
+        battle_data = {
+            "creator": self.creator,
+            "opponent": self.opponent,
+        }
+        self.client.login(username=self.creator.email, password='admin')
+        response = self.client.post('/battle/', battle_data)
+        import ipdb; ipdb.set_trace()
+        # response.context_data.view.refresh_from_db()
+        battle = Battle.objects.all()
+        self.assertEqual(battle, True)
--- a/backend/battle/tests.py
+++ b/backend/battle/tests.py
@@ -37,12 +37,10 @@ class BattleCreateViewTest(TestCase):
     
     def test_create_battle_successfully(self):
         battle_data = {
-            "creator": self.creator,
-            "opponent": self.opponent,
+            "creator": self.creator.id,
+            "opponent": self.opponent.id,
         }
         self.client.login(username=self.creator.email, password='admin')
         response = self.client.post('/battle/', battle_data)
-        import ipdb; ipdb.set_trace()
-        # response.context_data.view.refresh_from_db()
-        battle = Battle.objects.all()
-        self.assertEqual(battle, True)
+        battle = Battle.objects.filter(creator=self.creator, opponent=self.opponent)
+        self.assertTrue(battle)
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -6,8 +6,6 @@ from battle.battles.battle import get_winner_for
 
 
 class ListBattlesTest(TestCaseUtils):
-    def setUp(self):
-        super().setUp()
 
     def test_login_user_can_acess_battle_list(self):
         response = self.auth_client.get(reverse('battle_list'))
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -12,19 +12,59 @@ class ListBattlesTest(TestCase):
         self.user = User.objects.create(email='example@example.com')
         self.user.set_password('admin')
         self.user.save()
-    
+
     def test_login_user_can_acess_battle_list(self):
-        # import ipdb; ipdb.set_trace()
         self.client.login(username=self.user.email, password='admin')
-        response = self.client.get('/battle/list/')
+        response = self.client.get(reverse('battle_list'))
         self.assertEqual(response.status_code, 200)
-    
-    def test_data_returns_battle_ids(self):
+
+    def test_data_returns_empty_list(self):
         self.client.login(username=self.user.email, password='admin')
-        battles = baker.make('battle.Battle', creator=self.user, _quantity=2)
+        response = self.client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [])
+
+    def test_data_returns_one_battle_in_list(self):
+        self.client.login(username=self.user.email, password='admin')
+        battle = baker.make('battle.Battle', creator=self.user)
+        response = self.client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [battle])
+
+    def test_data_returns_few_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=4)
         response = self.client.get(reverse('battle_list'))
         response_qs = response.context_data.get('battle_list')
         self.assertCountEqual(battles, response_qs)
+
+    def test_data_returns_a_lot_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
+
+    def test_data_an_error_when_the_user_is_not_logged(self):
+        response = self.client.get(reverse('battle_list'))
+        self.assertEqual(response.status_code, 302)
+
+    def test_data_diff_to_verify_it_returns_exactly_len_of_battle_list(self):
+        self.client.login(username=self.user.email, password='admin')
+
+        baker.make('battle.Battle', creator=self.user, _quantity=10)
+        battles_initial = Battle.objects.filter(creator=self.user)
+        response_initial = self.client.get(reverse('battle_list'))
+        response_qs_initial = response_initial.context_data.get('battle_list')
+
+        self.assertCountEqual(battles_initial, response_qs_initial)
+
+        baker.make('battle.Battle', creator=self.user, _quantity=90)
+        battles_updated = Battle.objects.filter(creator=self.user)
+        response_updated = self.client.get(reverse('battle_list'))
+        response_qs_updated = response_updated.context_data.get('battle_list')
+
+        self.assertCountEqual(battles_updated, response_qs_updated)
 
 
 class BattleCreateViewTest(TestCase):
@@ -35,7 +75,7 @@ class BattleCreateViewTest(TestCase):
         self.creator.set_password('admin')
         self.creator.save()
         self.opponent = baker.make('users.User')
-    
+
     def test_create_battle_successfully(self):
         battle_data = {
             "creator": self.creator.id,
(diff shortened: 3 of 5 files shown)


backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
-# from django.test import TestCase
+from django.test import TestCase, Client
+from model_bakery import baker
+from users.models import User
+from battle.models import Battle
 
-# Create your tests here.
+'''
```
rsarai: Any reason for this to be commented? :eyes:

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+class ListBattlesTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+        self.battle = baker.make('battle.Battle')
+        self.user = User.objects.create(email='deborah.mendonca@vinta.com.br', password='admin')
```
rsarai: please use a fake email here

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        self.user.set_password('admin')
```
rsarai: I don't think you need to set the password here after the previous line.

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        self.user.save()
+    
+    def test_login_user_can_acess_battle_list(self):
+        # import ipdb; ipdb.set_trace()
```
rsarai: :eyes:

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        self.client.login(username=self.user.email, password='admin')
+        response = self.client.get('/battle/list/')
```
rsarai: You can and should use Django's `reverse('namespace')` to make the tests future proof

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        self.assertEqual(response.status_code, 200)
+    
+    def test_data_returns_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=2)
+        response = self.client.get('/battle/list/')
```
rsarai: Same as before

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
```
rsarai: You can complement this test by testing if the contents are the same

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+'''
+
+class BattleCreateViewTest(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.client = Client()
+        self.creator = User.objects.create(email='deborah.mendonca@vinta.com.br', password='admin')
+        self.creator.set_password('admin')
+        self.creator.save()
```
rsarai: same as before

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
-# from django.test import TestCase
+from django.test import TestCase, Client
+from model_bakery import baker
+from users.models import User
+from battle.models import Battle
 
-# Create your tests here.
+'''
+class ListBattlesTest(TestCase):
```
rsarai: When writing tests for things that return a collection, test:
  - Zero
  - One
  - Few
  - A lot
  - Exception
  - [The diff ](https://vinta.atlassian.net/wiki/spaces/development/blog/2021/07/14/2024079365/Testing+the+diff)

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+    
+    def test_data_returns_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=2)
+        response = self.client.get('/battle/list/')
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
+'''
+
+class BattleCreateViewTest(TestCase):
```
rsarai: Other useful test cases are test creator and opponent are the same and with an id that doesn't exist

backend/battle/tests.py
```diff
@@ -1,3 +1,46 @@
...
+        self.creator.save()
+        self.opponent = baker.make('users.User')
+    
+    def test_create_battle_successfully(self):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": self.opponent.id,
+        }
+        self.client.login(username=self.creator.email, password='admin')
+        response = self.client.post('/battle/', battle_data)
```
rsarai: You can and should use Django's reverse('namespace') to make the tests future proof

rsarai (changes requested):
Thank you for the small PR, left some comments

backend/battle/tests.py
DeborahAlmeida: thanks :)

backend/battle/tests.py
DeborahAlmeida: I tested, it doesn't work when I set the password in the previous line, but I get success when I call the set_password(). So, I'm going to delete the unnecessary `password='admin' `

backend/battle/tests.py
DeborahAlmeida: sorry, thanks for show me :)

backend/battle/tests.py
DeborahAlmeida: got it ;)

backend/battle/tests.py
DeborahAlmeida: thank you! :)

backend/battle/tests.py
DeborahAlmeida: nop, it's wrong. I'm going to delete it.

backend/battle/tests.py
DeborahAlmeida: Alright :)

backend/battle/tests.py
DeborahAlmeida: I searched about the assertCountEqual and I discovered this verifies if the lengths of the lists and the contents are the same. Please let me know if I understood wrong.

backend/battle/tests.py
DeborahAlmeida: great! Thanks :)

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+        self.client = Client()
+        self.battle = baker.make('battle.Battle')
+        self.user = User.objects.create(email='example@example.com')
+        self.user.set_password('admin')
+        self.user.save()
+
+    def test_login_user_can_acess_battle_list(self):
+        self.client.login(username=self.user.email, password='admin')
+        response = self.client.get(reverse('battle_list'))
+        self.assertEqual(response.status_code, 200)
```
rsarai: Awesome :rocket:

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+
+    def test_data_returns_empty_list(self):
```
rsarai: I think the naming of the tests here could be improved, the `data` is not making much sense. Maybe `test_view_returns...`

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+    def test_data_returns_a_lot_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
+
+    def test_data_an_error_when_the_user_is_not_logged(self):
+        response = self.client.get(reverse('battle_list'))
+        self.assertEqual(response.status_code, 302)
```
rsarai, line 42: This view is probably redirecting the user to the login view. You can assert that this is the case with `assertRedirects`

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+
+    def test_data_diff_to_verify_it_returns_exactly_len_of_battle_list(self):
+        self.client.login(username=self.user.email, password='admin')
```
rsarai: Since this is being used on almost all tests, you can move this to the `setUp`

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+            response.context_data['form'].errors['__all__'][0],
+            "ERROR: You can't challenge yourself.")
+
+    def test_if_returns_error_when_opponent_doesnt_exist(self):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": 100,
+        }
+        self.client.login(username=self.creator.email, password='admin')
+        with self.assertRaises(KeyError):
```
rsarai: When a view raises an exception it's bad for the user. I suggest you to fix the view to not raise an exception and raise a validation error instead

backend/battle/tests/test_view.py
```diff
@@ -1,9 +1,113 @@
...
+        self.client.post(reverse('battle'), battle_data)
+        battle = Battle.objects.filter(creator=self.creator, opponent=self.opponent)
+        self.assertTrue(battle)
+
+    def test_if_returns_error_when_creator_and_opponent_are_the_same_user(self):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": self.creator.email,
+        }
+        self.client.login(username=self.creator.email, password='admin')
```
rsarai: Noticed that you are using this in multiple views. A good approach here is to abstract this to a class that inherits from TestCase and already has a `self.auth_client`. Let me know if you need any help here

rsarai (changes requested):

backend/battle/tests/test_view.py
DeborahAlmeida: Thank you :)

backend/battle/tests/test_view.py
DeborahAlmeida: Great! Thanks

backend/battle/tests/test_view.py
DeborahAlmeida: Alright :)

backend/battle/tests/test_view.py
DeborahAlmeida: You're right, I'm going to fix it :)

backend/battle/tests/test_view.py
DeborahAlmeida: Good idea, thanks!!

backend/battle/tests/test_view.py
DeborahAlmeida: hmm, great! I'm going to use the auth_client you mentioned.

backend/battle/forms.py
```diff
@@ -46,6 +46,9 @@ def clean(self):
         if 'creator' not in cleaned_data:
             raise forms.ValidationError("ERROR: You need to be logged.")
 
+        if 'opponent' not in cleaned_data:
```
rsarai, line 49: I'm not sure, but I think this can be replaced by a `required=True` on the attribute definition (`opponent = forms.EmailField()`)

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
     def setUp(self):
         super().setUp()
-        self.creator, self.opponent = baker.make('users.User', _quantity=2)
-        battle = baker.make('battle.Battle', creator=self.creator, opponent=self.opponent)
-        self.team_creator = baker.make('battle.Team', battle=battle, trainer=self.creator)
+        self.battle = baker.make('battle.Battle', creator=self.user)
+        self.battle_from_user = Battle.objects.filter(creator=self.user)
+
+    def test_login_user_can_acess_battle_list(self):
+        response = self.auth_client.get(reverse('battle_list'))
```
rsarai, line 11: Much better!

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+        battle = Battle.objects.get(creator=self.user)
+        battle.delete()
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [])
+
+    def test_view_returns_one_battle_in_list(self):
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [self.battle_from_user][0])
```
rsarai: Why not do `[self.battle]`?

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
 
-class GetWinnerTest(TestCase):
+class ListBattlesTest(TestCaseUtils):
     def setUp(self):
         super().setUp()
-        self.creator, self.opponent = baker.make('users.User', _quantity=2)
-        battle = baker.make('battle.Battle', creator=self.creator, opponent=self.opponent)
-        self.team_creator = baker.make('battle.Team', battle=battle, trainer=self.creator)
+        self.battle = baker.make('battle.Battle', creator=self.user)
+        self.battle_from_user = Battle.objects.filter(creator=self.user)
```
rsarai: I'm seeing that you are using this to avoid making queries on tests. I would like to discourage this, especially because it's not clear that each test needs this specific query. Remember "Explicit is better than implicit". Another issue with this will be repeat for all tests even the ones that don't require this query

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+    def test_view_returns_one_battle_in_list(self):
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [self.battle_from_user][0])
+
+    def test_view_returns_few_battle_ids(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=4)
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(self.battle_from_user, response_qs)
```
rsarai: Please replace the `self.battle_from_user` with `Battle.objects.filter(creator=self.user)`

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+
+    def test_view_returns_a_lot_battle_ids(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(self.battle_from_user, response_qs)
```
rsarai: Same as before

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+
+    def test_view_an_error_when_the_user_is_not_logged(self):
+        self.auth_client.logout()
+        response = self.client.get(reverse('battle_list'))
+        self.assertRedirects(response, '/accounts/login/?next=/battle/list/')
```
rsarai, line 41: you can keep here the status_code verification;

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+        super().setUp()
+        self.opponent = baker.make('users.User')
+
+    def test_create_battle_successfully(self):
+        battle_data = {
+            "creator": self.user.id,
+            "opponent": self.opponent.email,
+        }
+        self.auth_client.post(reverse('battle'), battle_data)
+        battle = Battle.objects.filter(creator=self.user, opponent=self.opponent)
```
rsarai, line 71: I think this can be a `.get`

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
...
+        self.assertTrue(battle)
```
rsarai, line 72: Please check if the creator and opponent are the same as the ones you set, and that the winner is none

backend/battle/tests/test_view.py
```diff
@@ -1,15 +1,104 @@
-from django.test import TestCase
 from model_bakery import baker
-from battle.models import PokemonTeam
+from common.utils.tests import TestCaseUtils
```
rsarai, line 2: I didn't see this class, is it in another PR?

rsarai (changes requested):

backend/battle/forms.py
DeborahAlmeida: I tried this the `required=True` and it didn't any effect. So I kept this validation.

backend/battle/tests/test_view.py
DeborahAlmeida: That's true, thanks!!

backend/battle/tests/test_view.py
DeborahAlmeida: Thanks Rebeca! I'm going to fix it :D

backend/battle/tests/test_view.py
DeborahAlmeida: Great!

backend/battle/tests/test_view.py
DeborahAlmeida: Yup

backend/battle/tests/test_view.py
DeborahAlmeida: okay :)

backend/battle/tests/test_view.py
DeborahAlmeida: You're right!

backend/battle/tests/test_view.py
DeborahAlmeida: Thank you! But I have one question. When I filter by the creator and opponent I set, it doesn't do this? (talking about verifying the creator and opponent)
About the winner >> I'm going to do this :) thanks!!

backend/battle/tests/test_view.py
DeborahAlmeida: This already exists on the boilerplate :)

backend/battle/tests/test_view.py
DeborahAlmeida: Because I was filtering, but now as you mentioned, I changed to just get method

backend/battle/tests/test_view.py
rsarai: yes, ok I will let this go

rsarai (approved):

PR #29, opened 2021-07-24
Feature/test create team pokemon

DeborahAlmeida:
# Objective and Motivation
- Test if the team is created successfully
- Test error cases

### DIFF
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -73,7 +73,7 @@ class TeamForm(forms.ModelForm):
 
         pokemons_exist = verify_pokemon_exists_api([pokemon_1, pokemon_2, pokemon_3])
         if not pokemons_exist:
-            raise forms.ValidationError('ERROR: Type the correct pokemons name')
+            raise forms.ValidationError('ERROR: Type the correct pokemons name' )
 
         valid_pokemons = validate_sum_pokemons(
             [
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -59,17 +59,22 @@ class TeamForm(forms.ModelForm):
         pokemon_1 = self.cleaned_data.get('pokemon_1')
         pokemon_2 = self.cleaned_data.get('pokemon_2')
         pokemon_3 = self.cleaned_data.get('pokemon_3')
-        obj_battle = cleaned_data['battle']
         obj_trainer = cleaned_data['trainer']
-        position_pkn_1 = cleaned_data['position_pkn_1']
-        position_pkn_2 = cleaned_data['position_pkn_2']
-        position_pkn_3 = cleaned_data['position_pkn_3']
+
+        if 'battle' not in cleaned_data:
+            raise forms.ValidationError('ERROR: Select a valid battle')
+
+        obj_battle = cleaned_data['battle']
 
         if not pokemon_1 or not pokemon_2 or not pokemon_3:
             raise forms.ValidationError('ERROR: Select all pokemons')
 
-        if not position_pkn_1 or not position_pkn_2 or not position_pkn_3:
+        if ('position_pkn_1' or 'position_pkn_2' or 'position_pkn_3') not in cleaned_data:
             raise forms.ValidationError('ERROR: Select all positions')
+
+        position_pkn_1 = cleaned_data['position_pkn_1']
+        position_pkn_2 = cleaned_data['position_pkn_2']
+        position_pkn_3 = cleaned_data['position_pkn_3']
 
         pokemons_exist = verify_pokemon_exists_api([pokemon_1, pokemon_2, pokemon_3])
         if not pokemons_exist:
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -73,7 +73,7 @@ class TeamForm(forms.ModelForm):
 
         pokemons_exist = verify_pokemon_exists_api([pokemon_1, pokemon_2, pokemon_3])
         if not pokemons_exist:
-            raise forms.ValidationError('ERROR: Type the correct pokemons name' )
+            raise forms.ValidationError('ERROR: Type the correct pokemons name')
 
         valid_pokemons = validate_sum_pokemons(
             [
--- a/backend/battle/forms.py
+++ b/backend/battle/forms.py
@@ -73,7 +73,7 @@ class TeamForm(forms.ModelForm):
 
         pokemons_exist = verify_pokemon_exists_api([pokemon_1, pokemon_2, pokemon_3])
         if not pokemons_exist:
-            raise forms.ValidationError('ERROR: Type the correct pokemons name ')
+            raise forms.ValidationError('ERROR: Type the correct pokemons name')
 
         valid_pokemons = validate_sum_pokemons(
             [
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -27,7 +27,7 @@ class TeamViewTest(TestCaseUtils):
             reverse("team_create", kwargs={'pk': self.battle.id}), pokemons_data, follow=True)
         self.assertEqual(response.status_code, 200)
 
-        team_user = Team.objects.filter(battle=self.battle, trainer=self.user)
+        team_user = Team.objects.get(battle=self.battle, trainer=self.user)
         self.assertTrue(team_user)
 
         pokemons_on_team_user = []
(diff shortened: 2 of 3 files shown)


backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
+from django.test import TestCase, Client
+from model_bakery import baker
+from django.urls import reverse
+from battle.models import Team
+from users.models import User
+
+
+class AssociatePokemonsToBattleTest(TestCase):
```
rsarai: Is a good practice to put the name of the test the same of view plus `Test`

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
...
+        self.user = User.objects.create(email='example@example.com')
+        self.user.set_password('admin')
+        self.user.save()
+        self.battle = baker.make('battle.Battle', creator=self.user)
+        names = ['pikachu', 'bulbasaur', 'pidgeot', 'wrong_name']
+        for pokemon_name in names:
+            baker.make("pokemon.Pokemon", name=pokemon_name, attack=60, defense=15, hp=10)
+
+    def test_user_can_create_team_with_pokemons(self):
+        self.client.login(username=self.user.email, password='admin')
```
rsarai: I suggested on some other PR to extract this to an utils class

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
...
+            "pokemon_3": 'pidgeot',
+            "position_pkn_1": 1,
+            "position_pkn_2": 2,
+            "position_pkn_3": 3,
+        }
+        response = self.client.post(
+            reverse("team_create", kwargs={'pk': self.battle.id}), pokemons_data, follow=True)
+        self.assertEqual(response.status_code, 200)
+        team_user = Team.objects.filter(trainer=self.user)
+        self.assertTrue(team_user)
```
rsarai, line 191: Please assert that the content of the team is correct

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
...
+    def test_if_returns_error_when_missing_position(self):
+        self.client.login(username=self.user.email, password='admin')
+        pokemons_data = {
+            "battle": self.battle.id,
+            "trainer": self.user.id,
+            "pokemon_1": 'pikachu',
+            "pokemon_2": 'bulbasaur',
+            "pokemon_3": 'pidgeot',
+        }
+        with self.assertRaises(KeyError):
```
rsarai: Please make the order fields required so that the form raises a validation error

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
...
+            self.client.post(reverse(
+                "team_create", kwargs={'pk': self.battle.id}), pokemons_data, follow=True)
+
+    def test_if_returns_error_when_user_is_not_permited(self):
```
rsarai, line 238: awesome! :rocket:

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,123 @@
...
+            "pokemon_1": 'pikachu',
+            "pokemon_2": 'bulbasaur',
+            "pokemon_3": 'pidgeot',
+            "position_pkn_1": 1,
+            "position_pkn_2": 2,
+            "position_pkn_3": 3,
+        }
+        with self.assertRaises(KeyError):
+            self.client.post(reverse(
+                "team_create", kwargs={'pk': 100}), pokemons_data, follow=True)
```
rsarai: As I mentioned it's not good when a view raises an exception. We should add validation to the view level here to raise a 404 error

rsarai (changes requested):

backend/battle/tests/test_view.py
DeborahAlmeida: hmm great!

backend/battle/tests/test_view.py
DeborahAlmeida: I updated it in this PR, does it have some problem?

backend/battle/tests/test_view.py
DeborahAlmeida: okay :)

backend/battle/tests/test_view.py
DeborahAlmeida: Got it :)

backend/battle/tests/test_view.py
DeborahAlmeida: thanks :dancers:

backend/battle/tests/test_view.py
DeborahAlmeida: great! Thanks again :D

backend/battle/forms.py
```diff
@@ -59,21 +59,26 @@ def clean(self):
         pokemon_1 = self.cleaned_data.get('pokemon_1')
         pokemon_2 = self.cleaned_data.get('pokemon_2')
         pokemon_3 = self.cleaned_data.get('pokemon_3')
-        obj_battle = cleaned_data['battle']
         obj_trainer = cleaned_data['trainer']
-        position_pkn_1 = cleaned_data['position_pkn_1']
-        position_pkn_2 = cleaned_data['position_pkn_2']
-        position_pkn_3 = cleaned_data['position_pkn_3']
+
+        if 'battle' not in cleaned_data:
```
rsarai, line 102: If I'm not mistaken this can be made by adding `required=True` on the field definition

backend/battle/forms.py
```diff
@@ -59,21 +59,26 @@ def clean(self):
...
+            raise forms.ValidationError('ERROR: Select a valid battle')
+
+        obj_battle = cleaned_data['battle']
 
         if not pokemon_1 or not pokemon_2 or not pokemon_3:
             raise forms.ValidationError('ERROR: Select all pokemons')
 
-        if not position_pkn_1 or not position_pkn_2 or not position_pkn_3:
+        if ('position_pkn_1' or 'position_pkn_2' or 'position_pkn_3') not in cleaned_data:
```
rsarai, line 110: Same here

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,134 @@
...
+            "pokemon_3": 'pidgeot',
+            "position_pkn_1": 1,
+            "position_pkn_2": 2,
+            "position_pkn_3": 3,
+        }
+        response = self.auth_client.post(
+            reverse("team_create", kwargs={'pk': self.battle.id}), pokemons_data, follow=True)
+        self.assertEqual(response.status_code, 200)
+
+        team_user = Team.objects.filter(battle=self.battle, trainer=self.user)
```
rsarai: I think a `get` here would do the job

backend/battle/tests/test_view.py
```diff
@@ -0,0 +1,134 @@
...
+            pokemons_data['pokemon_3']
+        ]
+
+        all_pokemons_on_team = PokemonTeam.objects.filter(
+            team=team_user[0]).prefetch_related('pokemon')
+
+        for count in enumerate(pokemons_data_submited):
+            pokemons_on_team_user.append(all_pokemons_on_team[count[0]].pokemon.name)
+
+        self.assertEqual(pokemons_on_team_user, pokemons_data_submited)
```
rsarai: You can do here:

```
pokemons_data_submited = [
    pokemons_data['pokemon_1'],
    pokemons_data['pokemon_2'],
    pokemons_data['pokemon_3']
]
self.assertEqual(
    set(pokemons_data_submited), 
    set([p.name for p in team.pokemons.all()])
)
```

backend/battle/views.py
```diff
@@ -46,7 +46,14 @@ class TeamView(LoginRequiredMixin, CreateView):
     success_url = reverse_lazy("invite")
 
     def get_initial(self):
-        initial = {"battle": self.kwargs['pk'], "trainer": self.request.user.id}
+        battle_id = self.kwargs['pk']
+
+        try:
+            Battle.objects.get(id=battle_id)
```
rsarai: Nice, this can be streamlined by `get_object_or_404`

rsarai: If the battle doesn't exist we should return an error page. Some people discuss that the behavior should be different to avoid leaking information of the application but for this case let's return a 404

rsarai (changes requested):

backend/battle/forms.py
DeborahAlmeida: Here I verify if the battle is valid. The value of the battle always will be passed, so I verify if exists

backend/battle/forms.py
DeborahAlmeida: When I submit the form without pokemons value I already receive the message "The filed is required". But I wanted to create this validation to verify on test

backend/battle/tests/test_view.py
DeborahAlmeida: yup!

backend/battle/tests/test_view.py
DeborahAlmeida: Thanks! This will make easy many things!

backend/battle/views.py
DeborahAlmeida: Alright!

backend/battle/forms.py
rsarai: if this is already on the form, you shouldn't need to add this extra validation :thinking:

rsarai (approved):

PR #30, opened 2021-07-26
Feature/test send email invite

DeborahAlmeida:
## Objective and Motivation]
- ( using Mock )
- Poke API Integration test
- Email sending test
- Run battle result task test

### DIFF
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -1,8 +1,11 @@
 from django.test import TestCase, Client
+from django.conf import settings
+from unittest.mock import patch
 from model_bakery import baker
 from django.urls import reverse
 from users.models import User
 from battle.models import Battle
+from pokemon.helpers import get_pokemon_from_api
 
 
 class ListBattlesTest(TestCase):
@@ -105,3 +108,47 @@ class BattleCreateViewTest(TestCase):
         self.client.login(username=self.creator.email, password='admin')
         with self.assertRaises(KeyError):
             self.client.post(reverse('battle'), battle_data)
+
+    @patch("battle.battles.email.send_templated_mail")
+    def test_if_battle_invitation_email_is_sent(self, mock_templated_mail):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": self.opponent.id,
+        }
+
+        self.client.login(username=self.creator.email, password='admin')
+
+        self.client.post(reverse("battle"), battle_data)
+
+        battle = Battle.objects.filter(
+            creator=self.creator, opponent=self.opponent)
+
+        self.assertTrue(battle)
+
+        mock_templated_mail.assert_called_with(
+            template_name="invite_challenge",
+            from_email=settings.FROM_EMAIL,
+            recipient_list=[self.opponent.email],
+            context={
+                'creator': self.creator.email,
+            },
+        )
+
+
+class IntegrationPokeapiTest(TestCase):
+    def test_if_pokeapi_integration_returns_pokemon_data_sucessfully(self):
+        fake_json = {
+            "defense": 40,
+            "attack": 55,
+            "hp": 35,
+            "name": 'pikachu',
+            "img_url": 'https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/25.png',
+            "pokemon_id": 25,
+        }
+
+        with patch("pokemon.helpers.requests") as mock_get:
+            mock_get.return_value.json.return_value = fake_json
+
+        response = get_pokemon_from_api('pikachu')
+
+        self.assertEqual(response, fake_json)
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -216,7 +216,7 @@ class BattleCreateViewTest(TestCaseUtils):
         result_battle_mock.assert_called_with(battle)
 
 
-class IntegrationPokeapiTest(TestCaseUtils):
+class IntegrationPokeApiTest(TestCaseUtils):
 
     @patch('battle.battles.battle.get_pokemon_from_api')
     def test_if_pokeapi_integration_returns_pokemon_data_sucessfully(self, mock_get_pokemon):
@@ -262,7 +262,50 @@ class IntegrationPokeapiTest(TestCaseUtils):
             mock_get_pokemon.assert_called_with(pokemon)
             total_point_pokemons += pokemon_point
 
+        self.assertEqual(total_point_pokemons, 310 )
+
         is_valid_sum = validate_sum_pokemons(['pikachu', 'pidgey', 'bulbasaur'])
+        self.assertTrue(is_valid_sum)
+
+    @patch('battle.battles.battle.get_pokemon_from_api')
+    def test_if_function_valid_points_returns_corretly(self, mock_get_pokemon):
+        def side_effect_func(pokemon_name):
+            fake_json = 1
+            if pokemon_name == 'pikachu':
+                fake_json = {
+                    "defense": 40,
+                    "attack": 55,
+                    "hp": 35,
+                    "name": 'pikachu',
+                    "img_url": 'https://raw.githubusercontent.com'
+                            '/PokeAPI/sprites/master/sprites/pokemon/25.png',
+                    "pokemon_id": 25,
+                }
+            elif pokemon_name == 'pidgey':
+                fake_json = {
+                    "defense": 50,
+                    "attack": 25,
+                    "hp": 15,
+                    "name": 'pidgey',
+                    "img_url": 'https://raw.githubusercontent.com'
+                            '/PokeAPI/sprites/master/sprites/pokemon/25.png',
+                    "pokemon_id": 15,
+                }
+            elif pokemon_name == 'bulbasaur':
+                fake_json = {
+                    "defense": 30,
+                    "attack": 40,
+                    "hp": 20,
+                    "name": 'bulbasaur',
+                    "img_url": 'https://raw.githubusercontent.com'
+                            '/PokeAPI/sprites/master/sprites/pokemon/25.png',
+                    "pokemon_id": 10,
+                }
+            return fake_json
+        mock_get_pokemon.side_effect = side_effect_func
+
+        is_valid_sum = validate_sum_pokemons(['pikachu', 'pidgey', 'bulbasaur'])
+
         self.assertTrue(is_valid_sum)
 
 
@@ -299,11 +342,11 @@ class TeamViewTest(TestCaseUtils):
             "position_pkn_2": 2,
             "position_pkn_3": 3,
         }
-        response = self.auth_client.post(
+        self.auth_client.post(
             reverse("team_create", kwargs={'pk': battle.id}), pokemons_data, follow=True)
-        task_mock.return_value = 2
+
         task_mock.assert_called_with(battle.id)
-        # self.assertEqual(response, 1)
+
         self.assertTrue(Team.objects.get(battle=battle, trainer=self.user))
 
     @patch('battle.battles.email.send_templated_mail')
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -1,4 +1,3 @@
-from django.test import TestCase
 from unittest.mock import patch
 from model_bakery import baker
 from common.utils.tests import TestCaseUtils
@@ -228,6 +227,7 @@ class IntegrationPokeApiTest(TestCaseUtils):
         is_valid_sum = validate_sum_pokemons(['pikachu', 'pidgey', 'bulbasaur'])
 
         self.assertTrue(is_valid_sum)
+
 
 class TeamViewTest(TestCaseUtils):
     def setUp(self):
--- a/backend/battle/tests/test_view.py
+++ b/backend/battle/tests/test_view.py
@@ -1,198 +1,284 @@
-from django.test import TestCase
+from django.test import TestCase, RequestFactory
 from unittest.mock import patch
 from model_bakery import baker
 from common.utils.tests import TestCaseUtils
 from django.conf import settings
 from django.urls import reverse
 
-from battle.models import Battle, PokemonTeam
+from battle.models import Battle, PokemonTeam, Team
+from battle.views import TeamView
+from battle.forms import TeamForm
 from battle.battles.battle import get_winner_for, validate_sum_pokemons, sum_point_from_api
 from battle.tasks import run_battle_and_send_result_email
 
 
-class GetWinnerTest(TestCase):
-    def setUp(self):
-        super().setUp()
-        self.creator, self.opponent = baker.make('users.User', _quantity=2)
-        battle = baker.make('battle.Battle', creator=self.creator, opponent=self.opponent)
-        self.team_creator = baker.make('battle.Team', battle=battle, trainer=self.creator)
-        self.team_opponent = baker.make('battle.Team', battle=battle, trainer=self.opponent)
-        self.pokemon_1, self.pokemon_2, self.pokemon_3 = baker.make(
-            "pokemon.Pokemon",
-            attack=60, defense=45, hp=50, _quantity=3)
+# class GetWinnerTest(TestCase):
+#     def setUp(self):
+#         super().setUp()
+#         self.creator, self.opponent = baker.make('users.User', _quantity=2)
+#         battle = baker.make('battle.Battle', creator=self.creator, opponent=self.opponent)
+#         self.team_creator = baker.make('battle.Team', battle=battle, trainer=self.creator)
+#         self.team_opponent = baker.make('battle.Team', battle=battle, trainer=self.opponent)
+#         self.pokemon_1, self.pokemon_2, self.pokemon_3 = baker.make(
+#             "pokemon.Pokemon",
+#             attack=60, defense=45, hp=50, _quantity=3)
 
-    def test_verify_if_returns_creator_as_winner_when_draw(self):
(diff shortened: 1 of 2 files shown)


backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
 
         with self.assertRaises(IndexError):
             get_winner_for(self.team_creator, self.team_opponent)
+
+
+class ListBattlesTest(TestCase):
```
rsarai: I have the feeling that I already reviewed this? :thinking:

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+                team=team_opponent,
+                pokemon=pokemon,
+                order=count)
+
+        result_battle_mock.return_value = team_creator
+        run_battle_and_send_result_email.apply([battle.id])
+        result_battle_mock.assert_called_with(battle)
+
+
+class IntegrationPokeapiTest(TestCase):
```
rsarai: Nice, this is the way to do mocks. Now you need to mock `get_pokemon_from_api` on where it's used on the system.

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+class ListBattlesTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+        self.battle = baker.make('battle.Battle')
+        self.user = User.objects.create(email='example@example.com')
+        self.user.set_password('admin')
+        self.user.save()
+
+    def test_login_user_can_acess_battle_list(self):
+        self.client.login(username=self.user.email, password='admin')
```
rsarai: Since this is being used on a lot of classes, you should extract this to a separate class and create an `auth_client`

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        response = self.client.get(reverse('battle_list'))
+        self.assertEqual(response.status_code, 200)
+
+    def test_data_returns_empty_list(self):
```
rsarai: The naming could be improved here, I don't think `data` is a good fit here

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+    def test_data_returns_a_lot_battle_ids(self):
+        self.client.login(username=self.user.email, password='admin')
+        battles = baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(battles, response_qs)
+
+    def test_data_an_error_when_the_user_is_not_logged(self):
+        response = self.client.get(reverse('battle_list'))
+        self.assertEqual(response.status_code, 302)
```
rsarai: This is probably a redirect, you can assert that the redirect is correct with `assertRedirects`

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+
+    def test_create_battle_successfully(self):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": self.opponent.email,
+        }
+        self.client.login(username=self.creator.email, password='admin')
+        self.client.post(reverse('battle'), battle_data)
+        battle = Battle.objects.filter(creator=self.creator, opponent=self.opponent)
+        self.assertTrue(battle)
```
rsarai, line 148: Please assert the content of the battle as well

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+            response.context_data['form'].errors['__all__'][0],
+            "ERROR: You can't challenge yourself.")
+
+    def test_if_returns_error_when_opponent_doesnt_exist(self):
+        battle_data = {
+            "creator": self.creator.id,
+            "opponent": 100,
+        }
+        self.client.login(username=self.creator.email, password='admin')
+        with self.assertRaises(KeyError):
```
rsarai: I think I already commented on this, but we should not let views raises exceptions since it hurts the user experience, we should change to return a validation error

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        }
+
+        self.client.login(username=self.creator.email, password='admin')
+
+        self.client.post(reverse("battle"), battle_data)
+
+        battle = Battle.objects.filter(
+            creator=self.creator, opponent=self.opponent)
+
+        self.assertTrue(battle)
```
rsarai: this is not required since we are only testing the email and this is already being tested

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+
+        mock_templated_mail.assert_called_with(
+            template_name="invite_challenge",
+            from_email=settings.FROM_EMAIL,
+            recipient_list=[self.opponent.email],
+            context={
+                'creator': self.creator.email,
+            },
```
rsarai: Nice! :rocket:

backend/battle/tests/test_view.py
```diff
@@ -74,3 +81,178 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        )
+
+    @patch('battle.tasks.run_battle')
+    def test_if_run_battle_task(self, result_battle_mock):
```
rsarai: I think this test is misplaced. You should test if the `run_battle_and_send_result_email` is called properly and then test if the `run_battle_and_send_result_email` separately

rsarai (changes requested):

backend/battle/tests/test_view.py
DeborahAlmeida: yup, I merged from master branch because there were some conflicts

backend/battle/tests/test_view.py
DeborahAlmeida: yup, I fixed this on the other PR, sorry because there is on both PRs

DeborahAlmeida: I'm going to fix it on this PR as well

backend/battle/tests/test_view.py
DeborahAlmeida: thanks again :)

DeborahAlmeida (commented):
I revised by accident sorry

backend/battle/tests/test_view.py
DeborahAlmeida: got it

backend/battle/tests/test_view.py
DeborahAlmeida: yup

backend/battle/tests/test_view.py
DeborahAlmeida: hmm okay!

backend/battle/tests/test_view.py
DeborahAlmeida: thankss

backend/battle/tests/test_view.py
DeborahAlmeida: Great!

backend/battle/tests/test_view.py
DeborahAlmeida: Yup! Thanks :)

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
 
         with self.assertRaises(IndexError):
             get_winner_for(self.team_creator, self.team_opponent)
+
+
+class ListBattlesTest(TestCaseUtils):
+    def setUp(self):
+        super().setUp()
+        self.battle = baker.make('battle.Battle', creator=self.user)
+        self.battle_from_user = Battle.objects.filter(creator=self.user)
```
rsarai: Please avoid doing queries that will be used on asserts on the setup, add them explicitly on the `self.assertCountEqual`

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        battle = Battle.objects.get(creator=self.user)
+        battle.delete()
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [])
+
+    def test_view_returns_one_battle_in_list(self):
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [self.battle_from_user][0])
```
rsarai: Remove the `self.battle_from_user` by the actual query

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+
+    def test_view_returns_few_battle_ids(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=4)
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(self.battle_from_user, response_qs)
```
rsarai: same as above

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+
+    def test_view_returns_a_lot_battle_ids(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(self.battle_from_user, response_qs)
```
rsarai: same as above

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        self.auth_client.logout()
+        response = self.client.get(reverse('battle_list'))
+        self.assertRedirects(response, '/accounts/login/?next=/battle/list/')
+
+    def test_view_diff_to_verify_it_returns_exactly_len_of_battle_list(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=10)
+        response_initial = self.auth_client.get(reverse('battle_list'))
+        response_qs_initial = response_initial.context_data.get('battle_list')
+
+        self.assertCountEqual(self.battle_from_user, response_qs_initial)
```
rsarai: same as above

backend/battle/tests/test_view.py
rsarai: You can remove this test since it doesn't make sense to be on `BattleCreateViewTest` and it's still untouched

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+
+        for pokemon in ['pikachu', 'pidgey', 'bulbasaur']:
+            pokemon_point = sum_point_from_api(pokemon)
+            mock_get_pokemon.assert_called_with(pokemon)
+            total_point_pokemons += pokemon_point
+
+        self.assertEqual(total_point_pokemons, 310)
+
+        is_valid_sum = validate_sum_pokemons(['pikachu', 'pidgey', 'bulbasaur'])
+        self.assertTrue(is_valid_sum)
```
rsarai: This is already being tested on the following test, please remove it from here.

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+            "pokemon.Pokemon",
+            attack=60, defense=45, hp=50, _quantity=3)
+
+        pokemons = [pokemon_1, pokemon_2, pokemon_3]
+
+        for count, pokemon in enumerate(pokemons):
+            PokemonTeam.objects.create(
+                team=team_opponent,
+                pokemon=pokemon,
+                order=count)
```
rsarai: Shouldn't this be created by this view?

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
```
rsarai: Same as before

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        email_mock.assert_called_with(
+            template_name='results_battle',
+            from_email=settings.FROM_EMAIL,
+            recipient_list=[battle_updated.creator, battle_updated.opponent],
+            context={
+                'winner': battle_updated.winner,
+            },
+        )
+
+    def test_if_task_set_winner_correctly(self):
```
rsarai, line 438: This should in a separate test class since it doesn't relate to the `TeamViewTest`

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,343 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+                }
+            return fake_json
+        mock_get_pokemon.side_effect = side_effect_func
+
+        is_valid_sum = validate_sum_pokemons(['pikachu', 'pidgey', 'bulbasaur'])
+
+        self.assertTrue(is_valid_sum)
+
+
+class TeamViewTest(TestCaseUtils):
```
rsarai, line 232: I missed some tests here regarding the association of pokemon with the team. Is this is another PR?

rsarai (changes requested):

backend/battle/tests/test_view.py
DeborahAlmeida: Got it! :)

backend/battle/tests/test_view.py
DeborahAlmeida: Great!

backend/battle/tests/test_view.py
DeborahAlmeida: Okay!

backend/battle/tests/test_view.py
DeborahAlmeida: Yup :dancers:

backend/battle/tests/test_view.py
DeborahAlmeida: Alright :)

backend/battle/tests/test_view.py
DeborahAlmeida: Yup, I did it to understand how the mock works. Thanks :dancers:

backend/battle/tests/test_view.py
DeborahAlmeida: That's true. I will fix it :)

backend/battle/tests/test_view.py
DeborahAlmeida: Great!

backend/battle/tests/test_view.py
DeborahAlmeida: True!

backend/battle/tests/test_view.py
DeborahAlmeida: Yup! There's in another PR :)

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,271 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        battle = Battle.objects.get(creator=self.user)
+        battle.delete()
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [])
+
+    def test_view_returns_one_battle_in_list(self):
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(response_qs, [Battle.objects.filter(creator=self.user)][0])
```
rsarai: You can just use `Battle.objects.filter(creator=self.user)` or `list(Battle.objects.filter(creator=self.user))`

backend/battle/tests/test_view.py
```diff
@@ -74,3 +80,271 @@ def test_verify_if_not_returns_winner_missing_pokemons_teams(self):
...
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(Battle.objects.filter(creator=self.user), response_qs)
+
+    def test_view_returns_a_lot_battle_ids(self):
+        baker.make('battle.Battle', creator=self.user, _quantity=100)
+        response = self.auth_client.get(reverse('battle_list'))
+        response_qs = response.context_data.get('battle_list')
+        self.assertCountEqual(Battle.objects.filter(creator=self.user), response_qs)
+
+    def test_view_an_error_when_the_user_is_not_logged(self):
```
rsarai: A better name would be `test_view_returns_an_error_when_the_user_is_not_logged`

rsarai (approved):
Made some minor suggestions but looking good overall

backend/battle/tests/test_view.py
DeborahAlmeida: okay :)

PR #34, opened 2021-08-06
Feature/api battle create team

DeborahAlmeida:
## Objective and Motivation
- Make endpoint to create teams


backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
+from battle.battles.battle import validate_sum_pokemons, verify_pokemon_is_saved
+
+from pokemon.helpers import verify_pokemon_exists_api
+
+
+def verify_if_data_is_valid(data):  # pylint: disable=too-many-return-statements
```
rsarai: This is already an improvement on code duplication, however, this is still not optimal.
1. First, this function makes multiple validations, this is usually not a problem, but in this case, the name is too generic, by looking at the serializers and forms I don't know what is being validated. I advise you to split each validation into an individual function and calling them on the forms and serializers.
2. This function has multiple return types (boolean and strings), I strongly advise you to avoid this. This makes the behavior obscure, remember explicit is better than implicit.

backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
...
+    message_error = None
+
+    if 'battle' not in data:
+        message_error = 'ERROR: Select a valid battle'
+        return message_error
```
rsarai: This may be required on the form, but not on the serializer since you can set a field to be required

backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if data['trainer'] not in (data['battle'].creator, data['battle'].opponent):
+        message_error = ("ERROR: You do not have permission for this action.")
+        return message_error
```
rsarai: This validation on DRF should be a permission:  https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions

rsarai: For the form, this should be a mixin, you can use the LoginRequired mixin as example: https://docs.djangoproject.com/en/3.2/topics/auth/default/#the-loginrequired-mixin

backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
...
+        message_error = 'ERROR: Pokemons sum more than 600 points. Select again'
+        return message_error
+
+    if data['position_pkn_1'] in (data['position_pkn_2'], data['position_pkn_3']):
+        message_error = 'ERROR: You cannot add the same position'
+        return message_error
+
+    if data['position_pkn_2'] == data['position_pkn_3']:
+        message_error = 'ERROR: You cannot add the same position'
+        return message_error
```
rsarai: This validation can be improved. You can do this using sets: 
```
set(data['position_pkn_1'], data['position_pkn_2'], data['position_pkn_3']) != set(1,2,3)
```

backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
...
+
+    verify_pokemon_is_saved(
```
rsarai: This name is still misleading, isn't this saving the pokemon on the database if they not exist?

backend/services/create_team.py
```diff
@@ -0,0 +1,52 @@
+from battle.battles.battle import validate_sum_pokemons, verify_pokemon_is_saved
```
rsarai: The name of this folder is confusing. My advice for naming services is to use the name of the app they are in. In this case, `services/battle.py`

backend/battle/serializers.py
```diff
@@ -19,6 +28,98 @@ class Meta:
...
+        position_pkn_1 = self.validated_data['position_pkn_1']
+        position_pkn_2 = self.validated_data['position_pkn_2']
+        position_pkn_3 = self.validated_data['position_pkn_3']
+
+        validated_data.pop('pokemon_1')
+        validated_data.pop('pokemon_2')
+        validated_data.pop('pokemon_3')
+        validated_data.pop('position_pkn_1')
+        validated_data.pop('position_pkn_2')
+        validated_data.pop('position_pkn_3')
```
rsarai: hmmm, any specific reason for adding this?

backend/battle/serializers.py
```diff
@@ -19,6 +28,98 @@ class Meta:
...
+            "pokemon_3"
+        )
+
+    def validate(self, attrs):
+        valid = verify_if_data_is_valid(attrs)
+        if valid is not True:
+            raise serializers.ValidationError(valid)
+        return attrs
+
+    def create(self, validated_data):
```
rsarai, line 109: I wonder how are you testing that this is working?

rsarai (changes requested):

backend/battle/serializers.py
```diff
@@ -19,6 +28,98 @@ class Meta:
...
+            team=instance,
+            pokemon=pokemon_3_object,
+            order=position_pkn_3)
+        try:
+            Team.objects.get(
+                trainer=validated_data['battle'].creator,
+                battle=validated_data['battle'])
+            Team.objects.get(
+                trainer=validated_data['battle'].opponent,
+                battle=validated_data['battle'])
```
rsarai: hmmmm, why is this required?

backend/services/create_team.py
DeborahAlmeida: hmm great!

backend/services/create_team.py
DeborahAlmeida: yes, I will improve this name function :D

backend/services/create_team.py
DeborahAlmeida: thanks :)

backend/battle/serializers.py
DeborahAlmeida: What do you mean?

backend/battle/serializers.py
DeborahAlmeida: I verify if exists the creator and opponent's teams

backend/services/create_team.py
DeborahAlmeida: thanks for all the support :dancers:

backend/services/create_team.py
DeborahAlmeida: thank you :D

backend/battle/battles/battle.py
```diff
@@ -53,7 +53,7 @@ def validate_sum_pokemons(pokemons):
     return total_points <= 600
 
 
-def verify_pokemon_is_saved(pokemons):
+def save_pokemons_if_they_doenst_exist(pokemons):
```
rsarai: Typo: `save_pokemons_if_they_dont_exist`

backend/battle/forms.py
```diff
@@ -2,17 +2,22 @@
 from django.contrib.auth.forms import UserCreationForm
 
 from battle.models import Battle, PokemonTeam, Team
-from battle.battles.battle import validate_sum_pokemons, verify_pokemon_is_saved
+from battle.battles.battle import save_pokemons_if_they_doenst_exist
```
rsarai: This namespace is weird: `battle.battles.battle`. I'm not gonna stress about this but we should avoid this going forward

backend/battle/forms.py
```diff
@@ -2,17 +2,22 @@
...
+from services.battle import (
     validate_if_creator_and_opponent_has_different_contenders,
-    fetch_opponent_or_create_if_doenst_exist, send_invite_email_or_create_password_email)
+    fetch_opponent_or_create_if_doenst_exist,
+    send_invite_email_or_create_password_email,
+    verify_pokemons_fields_has_missing_values,
+    verify_fields_has_wrong_pokemon_name,
+    verify_positions_fields_has_missing_values,
+    verify_pokemon_sum_is_valid,
+    verify_positions_fields_has_duplicate_values,
```
rsarai: I think all these helpers that start with `verify` should start with `has` or `is`

backend/battle/serializers.py
```diff
@@ -19,6 +35,109 @@ class Meta:
...
+        label="Position",
+        required=False)
+    position_pkn_2 = serializers.ChoiceField(
+        choices=POSITION_CHOICES,
+        label="Position",
+        required=False)
+    position_pkn_3 = serializers.ChoiceField(
+        choices=POSITION_CHOICES,
+        label="Position",
+        required=False)
```
rsarai, line 58: These fields should be required no?

backend/battle/serializers.py
```diff
@@ -19,6 +35,109 @@ class Meta:
...
+
+        valid = verify_positions_fields_has_duplicate_values(attrs)
+        if valid is not True:
+            raise serializers.ValidationError(valid)
+
+        save_pokemons_if_they_doenst_exist(
+            [
+                attrs['pokemon_1'],
+                attrs['pokemon_2'],
+                attrs['pokemon_3']])
```
rsarai: Indentation is a bit weird here.

backend/battle/serializers.py
```diff
@@ -19,6 +35,109 @@ class Meta:
...
+            team=instance,
+            pokemon=pokemon_2_object,
+            order=position_pkn_2)
+
+        PokemonTeam.objects.create(
+            team=instance,
+            pokemon=pokemon_3_object,
+            order=position_pkn_3)
+
+        if len(list(validated_data['battle'].teams.all())) == 2:
```
rsarai: Do you need the `list` here?

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
...
+            email_template_name='registration/guest_email.html',
+            from_email=settings.FROM_EMAIL,)
+    else:
+        send_invite_email(opponent, creator)
+
+
+def verify_pokemons_fields_has_missing_values(data):
+    if ('pokemon_1' or 'pokemon_2' or 'pokemon_3') not in data:
+        message_error = 'ERROR: Select all pokemons'
+        return message_error
```
rsarai: As I mentioned before, please avoid creating functions that return missed types, in this case, boolean and string. The error messages should be in the form and in the serializer

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
...
+    return True
+
+
+def verify_positions_fields_has_missing_values(data):
+    if ('position_pkn_1' or 'position_pkn_2' or 'position_pkn_3') not in data:
+        message_error = 'ERROR: Select all positions'
+        return message_error
```
rsarai: Same as before

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
...
+    return True
+
+
+def verify_fields_has_wrong_pokemon_name(data):
+    pokemons_exist = verify_pokemon_exists_api(
+        [data['pokemon_1'], data['pokemon_2'], data['pokemon_3']])
+
+    if not pokemons_exist:
+        message_error = 'ERROR: Type the correct pokemons name'
+        return message_error
```
rsarai: Same as before

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
...
+    return True
+
+
+def verify_pokemon_sum_is_valid(data):
+    valid_pokemons = validate_sum_pokemons(
+        [data['pokemon_1'], data['pokemon_2'], data['pokemon_3']])
+
+    if not valid_pokemons:
+        message_error = 'ERROR: Pokemons sum more than 600 points. Select again'
+        return message_error
```
rsarai: Same as before

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def verify_positions_fields_has_duplicate_values(data):
+    positions = {
+        int(data['position_pkn_1']),
+        int(data['position_pkn_2']),
+        int(data['position_pkn_3'])}
+    if positions != {1, 2, 3}:
+        message_error = 'ERROR: You cannot add the same position'
+        return message_error
```
rsarai: Same as before

backend/services/battle.py
```diff
@@ -0,0 +1,85 @@
+from users.models import User
```
rsarai, line 1: As I mentioned before all helpers that start with `verify` should start with `has` or `is`

rsarai (changes requested):

backend/battle/forms.py
DeborahAlmeida: okay, thanks for alert me

backend/battle/forms.py
DeborahAlmeida: hmm okay! I misunderstood, thanks!

backend/battle/serializers.py
DeborahAlmeida: true!

DeborahAlmeida: Actually, if I set true, it returns an error, so I need to set False, and I verify if the field is empty on serializer

backend/battle/serializers.py
DeborahAlmeida: Okay, is this final, right?

backend/battle/serializers.py
DeborahAlmeida: actually, I don't, thanks!

backend/services/battle.py
DeborahAlmeida: hmm interesting, thanks!

backend/services/battle.py
DeborahAlmeida: thank you :)

PR #38, opened 2021-08-19
create endpoint to get current user

DeborahAlmeida:
## Objective and Motivation
- Create an endpoint to get the current user
- Create URL to the new API feature
- Create tests to the new endpoint


backend/battle/urls_api.py
```diff
@@ -6,4 +6,5 @@
     path("battle/<int:pk>/", endpoints.BattleDetail.as_view(), name="battle-detail"),
     path("battle/create/", endpoints.BattleCreate.as_view(), name="battle-create"),
     path("team/create/", endpoints.TeamCreate.as_view(), name="team-create"),
+    path("user/", endpoints.LoggedUser.as_view(), name="logged-user"),
```
rsarai: Leaving this one comment since I already reviewed this in another PR: I think this should be prefixed with `api/` and the name of View could be improved, I suggest: `CurrentUserAPI`

DeborahAlmeida: Thanks, Rebeca :) In the `pokebattle/urls.py` I set the prefix to all the URLs related to `urls_api`

DeborahAlmeida: like here 
image

rsarai:
@DeborahAlmeida Is this PR required? I think we can close this one since I already reviewed this in another place

PR #42, opened 2021-09-02
implement normalizr

DeborahAlmeida:
## Objective and Movitation
- Implement normalizr to set up battle data


.prospector.yaml
```diff
@@ -34,6 +34,7 @@ pylint:
     - no-self-use
     - no-member
     - too-many-ancestors
+    - django-not-configured
```
rsarai, line 37: What's the reason for this?

frontend/js/reducers/battleReducer.js
```diff
@@ -15,3 +15,5 @@ export const battleReducer = (state = initialState, action) => {
       return state;
   }
 };
+
+// const battles = new schema.Entity('');
```
rsarai: Is this supposed to be here?

frontend/js/utils/schema.js
```diff
@@ -0,0 +1,14 @@
+import { schema } from 'normalizr';
+
+export const user = new schema.Entity('user');
```
rsarai: I suggest you to add to all the schemas the `Schema` suffix

frontend/js/utils/schema.js
```diff
@@ -0,0 +1,14 @@
...
+export const pokemon = new schema.Entity('pokemon');
+
+export const battle = new schema.Entity('battle', {
+  creator: user,
+  opponent: user,
+  winner: user,
+  teams: [{ trainer: user, pokemons: [pokemon] }],
+});
+
+export const battleList = new schema.Array(battle);
```
rsarai: I think `battles` is better

frontend/js/pages/BattleDetail.js
```diff
@@ -13,34 +13,46 @@ function BattleDetail(props) {
   const { id } = useParams();
   const { battle } = props.battle;
   const { user } = props.user;
-
+  const fetchedBattle = props.location.query;
```
rsarai: where this is coming from? also what is this?

frontend/js/pages/BattleDetail.js
```diff
@@ -13,34 +13,46 @@ function BattleDetail(props) {
...
   useEffect(() => {
     if (!user) {
       props.getCurrentUser();
     }
-    props.getBattle(id);
+    if (!fetchedBattle) {
+      props.getBattle(id);
+    }
```
rsarai: Put this another `useEffect`, the intention here is to separate concerns into different hooks: https://pt-br.reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns

frontend/js/pages/BattleList.js
```diff
@@ -27,6 +27,10 @@ function BattleList(props) {
       </div>
     );
   }
+  const battleArray = Object.values(battles.entities.battle);
+  const pokemons = battles.entities.pokemon;
+  const battleUsers = battles.entities.user;
```
rsarai: Note, that you don't need to do this inside the component, you should do this inside `mapStateToProps`

rsarai: Also, while you are there, you can denormalize your data to make the component life easier and avoid names such as `battleUsers` : https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#denormalizeinput-schema-entities

frontend/js/pages/BattleList.js
```diff
@@ -38,25 +42,37 @@ function BattleList(props) {
...
-          {battles.map((battle) =>
+          {battleArray.map((battle) =>
             battle.winner ? (
               <li key={battle.id} className="item">
-                <a className="battle_settled" href={Urls.battle_detail_v2(battle.id)}>
+                <Link
+                  className="battle_settled"
+                  to={{
+                    pathname: Urls.battle_detail_v2(battle.id),
+                    query: { battle, pokemons, battleUsers },
```
rsarai: I see here what you were doing before, an ingenious solution, however, we make it better and simple! Notice, you are persisting to the location information you already have available on the store, you should in your battle detail get this information from the store using `mapStateToProps` and `denormalize` as I mentioned before

frontend/js/pages/BattleList.js
```diff
@@ -38,25 +42,37 @@ function BattleList(props) {
...
-          {battles.map((battle) =>
+          {battleArray.map((battle) =>
             !battle.winner ? (
               <li key={battle.id} className="item">
-                <a className="battle_ongoing" href={Urls.battle_detail_v2(battle.id)}>
+                <Link
+                  className="battle_ongoing"
+                  to={{
+                    pathname: Urls.battle_detail_v2(battle.id),
+                    query: { battle, pokemons, battleUsers },
```
rsarai: Same as before

frontend/js/pages/BattleList.js
```diff
@@ -38,25 +42,37 @@ function BattleList(props) {
...
+                </Link>
               </li>
             ) : null
           )}
         </div>
 
         <div className="your_opponent">
           <h3>On goind Battles</h3>
-          {battles.map((battle) =>
+          {battleArray.map((battle) =>
```
rsarai: I think the naming change here is unnecessary

frontend/js/pages/BattleDetail.js
rsarai: This should be removed, please get the necessary information from the store

frontend/js/pages/BattleDetail.js
rsarai: On the line of my previous comment, this should be removed

frontend/js/pages/BattleDetail.js
```diff
@@ -13,34 +13,46 @@ function BattleDetail(props) {
...
-
+  let pokemonsList = null;
+  const battleDetail = fetchedBattle ? fetchedBattle.battle : battle.entities.battle[id];
+  const trainerDetail = fetchedBattle ? fetchedBattle.battleUsers : battle.entities.user;
+  if (fetchedBattle) {
+    pokemonsList = fetchedBattle.pokemons ? fetchedBattle.pokemons : null;
+  } else if (battle.entities) {
+    pokemonsList = battle.entities.pokemon ? battle.entities.pokemon : null;
+  }
+  const { current, other } = orderTeamsByCurrentUser(battleDetail, user, trainerDetail);
```
rsarai: This logic is not 100%, but the update to be made here is to move all the logic to `mapStateToProps` and use denormalize to get the data in the right format for the component

frontend/js/components/CardTeam.js
```diff
@@ -2,11 +2,11 @@ import React from 'react';
 
 import CardPokemon from 'components/CardPokemon';
 
-function CardTeam({ pokemons }) {
+function CardTeam(props) {
```
rsarai: Based on my previous comments this will need to be updated

rsarai (changes requested):

frontend/js/reducers/battleReducer.js
DeborahAlmeida: nop, thanks!

frontend/js/utils/schema.js
DeborahAlmeida: Thanks!

frontend/js/utils/schema.js
DeborahAlmeida: yup!

frontend/js/pages/BattleDetail.js
DeborahAlmeida: Thanks for the advice! I'll fix this :)

frontend/js/pages/BattleList.js
DeborahAlmeida: Thank you!

frontend/js/components/CardTeam.js
DeborahAlmeida: Thanks!

frontend/js/pages/BattleList.js
DeborahAlmeida: Alright! :)

frontend/js/pages/BattleDetail.js
DeborahAlmeida: Thanks!

.prospector.yaml
rsarai: As I mentioned please check the version of the prospector, this could have broken the prospector, this is not something that you should ignore

frontend/js/pages/BattleDetail.js
```diff
@@ -1,46 +1,54 @@
...
+    if (!battles) {
+      props.getBattle(id);
+    }
+  }, []);
 
-  const { current, other } = orderTeamsByCurrentUser(battle, user);
+  if (!battle && !battles) {
+    return 'loading';
+  }
+  const denormalizedData = denormalizeBattleData(id, battles, battle);
```
rsarai: This name could be improved

frontend/js/pages/BattleDetail.js
```diff
@@ -71,8 +75,8 @@ function BattleDetail(props) {
 }
 
 const mapStateToProps = (store) => ({
-  battle: store.battle,
-  user: store.user,
+  battle: store.battle.battle,
+  user: store.user.user,
```
rsarai: Is there a way to remove this nesting behavior? If is to difficult please use lodash: https://lodash.com/docs/4.17.15#get

frontend/js/pages/BattleList.js
```diff
@@ -66,11 +69,20 @@ function BattleList(props) {
...
-const mapStateToProps = (store) => ({
-  battles: store.battle,
-  user: store.user,
-});
-
+function mapStateToProps(store) {
+  let battles = null;
+  if (store.battle.battles) {
+    battles = denormalize(
+      store.battle.battles.result,
```
rsarai: same here

frontend/js/pages/BattleList.js
```diff
@@ -66,11 +69,20 @@ function BattleList(props) {
...
+      battlesSchema,
+      store.battle.battles.entities
```
rsarai: same here

frontend/js/pages/BattleList.js
```diff
@@ -66,11 +69,20 @@ function BattleList(props) {
...
+    );
+  }
+  return {
+    battles,
+    user: store.user.user,
```
rsarai: same here

frontend/js/pages/BattleDetail.js
rsarai: Check for changes on the reducers

rsarai (changes requested):

.prospector.yaml
DeborahAlmeida: Thank you!

frontend/js/pages/BattleDetail.js
DeborahAlmeida: great!

frontend/js/pages/BattleDetail.js
DeborahAlmeida: All right! I'll improve it thanks :dancing_women:

frontend/js/pages/BattleList.js
```diff
@@ -1,75 +1,93 @@
...
       </ul>
     </div>
   );
 }
 
-const mapStateToProps = (store) => ({
-  battles: store.battle,
-  user: store.user,
-});
+function mapStateToProps(store) {
```
joaopslins, line 71:
```suggestion
function mapStateToProps(state) {
```

